PR #9, opened 2019-03-17
Add .gitignore and ignore target/

necaris:


dlrobertson:
Should we just snag [GitHub's Rust .gitignore](https://github.com/github/gitignore/blob/master/Rust.gitignore)?

necaris:
@dlrobertson good call, updated to use that instead

PR #10, opened 2019-03-17
Add manpage fixtures

### LINKED ISSUE #7: Add examples of man pages as input
manpages.ubuntu.com seems like a good source of man pages as input, running the existing [`deroff`](https://repology.org/project/deroff/versions) over some of them as input could provide fixtures for us to test the `deroff.rs` implementation.

necaris:
Closes #7 

Add a couple of existing man pages as fixtures for integration tests, as well as "deroffed" versions to serve as intermediary tests for `deroff.rs` when it gets to that level of completion.

Included `find (1)` because it's a large, well-documented command and so I assume the man page will be equally thorough, and a good test case; and the sole manpage I could find in the `fr.ISO8859-1` section as a test of multi-lingual capabilities.

(Since the source code of `deroff.py` includes some assumptions that input text is `latin1`-encoded, I thought that was safest. And the French page was chosen because I read enough French to know if the characters have been [mojibaked](https://en.wikipedia.org/wiki/Mojibake))

These were both downloaded from http://manpages.ubuntu.com/manpages/cosmic and run through [this current master version of `deroff.py`](https://github.com/fish-shell/fish-shell/blob/e7bfd1d71ca54df726a4f1ea14bd6b0957b75752/share/tools/deroff.py).


dlrobertson:
Nice! Are there some small man pages that we might be able to use in a unit test?

necaris:
@dlrobertson good idea! I added the man page for `docker-rmi`, which is about the smallest one I could find that had more than one actual option.

scooter-dangle:
Would you mind also running these through the original completion generator utility so they could also act as fixtures for the utility as well? (Lemme know if you don't have time—I can run them in a separate PR)

necaris:
No worries, I'll just run them through real quick :-)

On Sun, Mar 17, 2019 at 11:22 PM Scott Steele <notifications@github.com> wrote:
>
> Would you mind also running these through the original completion generator utility so they could also act as fixtures for the utility as well? (Lemme know if you don't have time—I can run them in a separate PR)
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.



-- 
Rami Chowdhury
"A mind all logic is like a knife all blade - it makes the hand bleed
that uses it." -- Rabindranath Tagore
+1-408-372-6496 / +44-7581-430-517 / +88-0189-245544

necaris:
@scooter-dangle @dlrobertson FYI, ran the script all the way through and added the `.fish` files. I also replaced `find` with `mlterm`, because it seemed that `find`'s documentation didn't quite conform to the expected shape, and for an initial batch of tests I figured let's stick with the happy path.

PR #12, opened 2019-03-18
Feature/is option

AlanSwenson:
Adds 2 is_option functions. Very similar to could_be description functions.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -790,9 +790,16 @@ struct TypeDeroff;
 //     def is_option(self, line):
 //         return line.startswith('-')
 
+#[test]
+fn test_is_option() {
+    assert!(!TypeDeroff::is_option("Not an Option"));
+    assert!(TypeDeroff::is_option("-Is an Option"));
+    assert!(!TypeDeroff::is_option(""));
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -717,10 +717,15 @@ impl TypeDarwin {
 
 //     def is_option(self, line):
 //         return line.startswith('.It Fl')
-
+#[test]
+fn test_TypeDarwin_is_option() {
+    assert!(!TypeDarwin::is_option("Not an Option"));
+    assert!(TypeDarwin::is_option(".It Fl Is an Option"));
+    assert!(!TypeDarwin::is_option(""));
+}
```
scooter-dangle: Formatting nit. Could you add a blank line between the test function and the `impl`?

AlanSwenson:
@scooter-dangle blank line added

src/main.rs
```diff
@@ -717,10 +717,16 @@ impl TypeDarwin {
 
 //     def is_option(self, line):
```
dlrobertson: This can be removed in a follow-up PR

PR #13, opened 2019-03-18
add function groff_replace_escapes and tests

AlanSwenson:


src/main.rs
```diff
@@ -708,10 +708,25 @@ impl TypeDarwin {
 //         line = line.replace('\\ ', ' ')
 //         line = line.replace('\& ', '')
 //         return line
+#[test]
+fn test_TypeDarwin_groff_replace_escapes() {
+    // tests for expected replacements
+    assert!(TypeDarwin::groff_replace_escapes(".Nm") == "CMDNAME");
+    assert!(TypeDarwin::groff_replace_escapes(r"\\ ") == " ");
```
scooter-dangle: Hmm. Not sure that this is correct. This string is different from the one in Python. Output should be `r"\ "`. In Python the `\` is acting as an escape, but since you're using `r` string prefix in Rust, you have two `\` in the string.

```rust
// Rust
assert_eq!(r"\\ ".len(), 3);
```

```python
# Python
assert(len('\\ ') == 2)
```

src/main.rs
```diff
@@ -708,10 +708,25 @@ impl TypeDarwin {
...
+    assert!(TypeDarwin::groff_replace_escapes(r"\ ") == r"\ ");
+    assert!(TypeDarwin::groff_replace_escapes(r"\\") == r"\\");
+    assert!(TypeDarwin::groff_replace_escapes(r"& ") == "& ");
+    assert!(TypeDarwin::groff_replace_escapes(r"\&") == r"\&");
+}
 
 impl TypeDarwin {
     fn groff_replace_escapes(line: &str) -> String {
-        unimplemented!()
+        line.replace(".Nm", "CMDNAME")
```
dlrobertson: Can you remove the commented python stub when adding a function? It makes reviewing easier

AlanSwenson:
@scooter-dangle I believe I have fixed this function, thanks again
@dlrobertson python stub removed from this pull

PR #14, opened 2019-03-18
g_spec_specletter / g_spec functions

necaris:


src/deroff.rs
```diff
@@ -1,352 +1,277 @@
...
-//     }
+struct Deroffer {}
+
+impl Deroffer {
+    // for the moment, return small strings, until we figure out what
+    // it should really be doing
+    fn g_specs_specletter(key: &str) -> Option<&'static str> {
+        Some(match key {
+            // Output composed latin1 letters
+            "-D" => &"\u{00d0}",
```
scooter-dangle: Ah...kk...played with Python2 vs Python3 on the command line. Makes sense now for these to be character codes rather than raw byte codes. Didn't realize that the Python 2 vs 3 encoding difference would apply to string escape sequences (e.g., `'\320'` as in this case).

:+1:

src/deroff.rs
```diff
@@ -1,352 +1,277 @@
...
+            "rq" => &"”",
+            "Bq" => &",,",
+            "oq" => &"`",
+            "cq" => &"'",
+            "aq" => &"'",
+            "dq" => &"\"",
+            "or" => &"|",
+            "at" => &"@",
+            "sh" => &"#",
+            "Eu" => &"\u{00a4}",
```
scooter-dangle: Hmm. Might either need to open a separate issue. My assumption would be that this should be `€`, but in Rust that would be `"\u{20ac}"`. (In the interest of the first port being naïve, an issue to revisit seems fine.)

src/deroff.rs
necaris: That's a good point, I'll include it in this PR :-)

src/deroff.rs
necaris: I've gone ahead and used the Euro symbol, but opened https://github.com/rust-dc/fish-manpage-completions/issues/16 for us to follow up later.

PR #17, opened 2019-03-18
feat: Add basic in/out mode to program

### LINKED ISSUE #1: Add flag/option for simple read from stdin, print to stdout
Should be easier to grok what program is doing and what general point is if I can see simple input/output behavior without having to attempt to do a bulk completion generation job.

scooter-dangle:
Running the program with only the `--stdout` flag and no file arguments
will cause the program to read a manpage from stdin and print the
resulting completions to stdout.

Fixes #1


src/main.rs
```diff
@@ -380,8 +378,17 @@ trait ManParser {
 //         else:
 //             return True
 
+#[derive(Copy, Clone, Debug)]
 struct Type1;
-// TODO impl ManParser for Type1
+impl ManParser for Type1 {
```
dlrobertson: nit: space between the `impl` and `struct`.

src/main.rs
```diff
@@ -1085,6 +1137,49 @@ fn file_is_overwritable(path: &Path) -> bool {
...
+        Box::new(Type1),
+        Box::new(Type2),
+        Box::new(Type3),
+        Box::new(Type4),
+        Box::new(TypeDarwin),
+        Box::new(TypeDeroff),
+    ]
+}
+
+use std::io::{Read, Write};
```
dlrobertson: Why was this moved here?

scooter-dangle: No great reason. Needed `Write` for the function below and was resolving a warning for it being `use`d twice.

I've amended main.rs to move all the root-level `use` statements to the top of the file.

PR #22, opened 2019-03-20
feat: add function TypeDarwin::trim_groff

AlanSwenson:
This one was a bit trickier with all the Regex so please double check my work on it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -716,23 +716,32 @@ impl ManParser for TypeDarwin {
     }
 }
 
-//     def trim_groff(self, line):
-//         # Remove initial period
-//         if line.startswith('.'):
-//             line = line[1:]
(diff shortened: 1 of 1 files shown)


scooter-dangle:
Note: See Zulip discussion on this PR starting [here](https://rust-dc.zulipchat.com/#narrow/stream/183309-general/topic/Fish-shell/near/161370993).

AlanSwenson:
@scooter-dangle new version up with all the changes discussed on zulip

scooter-dangle:
This is great! Sorry for dropping the ball on reviewing!

I'll change the `assert!(X == Y)`s to `assert_eq!(X, Y)`s and then merge. (When it's possible to use, the latter construction gives slightly better error messages when the assertion fails.)

PR #23, opened 2019-03-21
Feature/count argument dashes

AlanSwenson:
Thanks for the help on this on @scooter-dangle .

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -745,21 +745,54 @@ impl TypeDarwin {
     }
 }
 
-//     def count_argument_dashes(self, line):
-//         # Determine how many dashes the line has using the following regex hack
-//         # Look for the start of a line, followed by a dot, then a sequence of
-//         # one or more dashes ('Fl')
-//         result = 0
-//         if line.startswith('.'):
-//             line = line[4:]
-//             while line.startswith('Fl '):
-//                 result = result + 1
-//                 line = line[3:]
-//         return result
+#[test]
+fn test_replace_all() {
+    let (string, num) = replace_all("");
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -745,21 +745,54 @@ impl TypeDarwin {
...
+    let (string, num) = replace_all(".Fl Fl Fl jkl");
+    assert_eq!(string, "jkl");
+    assert_eq!(num, 2);
+}
+
+use std::borrow::Cow;
+fn replace_all(line: &str) -> (Cow<str>, u32) {
+    let mut result = 0;
+    (
+        regex!(r"^(?:\....)((?:[F][l]\s)*)").replace(&line, |captures: &regex::Captures| {
```
scooter-dangle:
```suggestion
        regex!(r"^(?:\....)((?:Fl\s)*)").replace(&line, |captures: &regex::Captures| {
```

src/main.rs
```diff
@@ -745,21 +745,54 @@ impl TypeDarwin {
...
+            result = (captures[1].len() / 3) as u32; // Divide by 3 since there are 3 bytes per `Fl\s` pattern
+            ""
+        }),
+        result,
+    )
+}
+
+#[test]
+fn test_TypeDarwin_count_argument_dashes() {
+    assert!(TypeDarwin::count_argument_dashes(".Fl Fl xx") == 1);
```
scooter-dangle: Would be a little more idiomatic to do `assert_eq!(X, Y)` rather than `assert!(X == Y)` for this and the following lines.

PR #24, opened 2019-04-11
feat: implement ManParser::is_my_type

scooter-dangle:


src/main.rs
```diff
@@ -1187,31 +1143,102 @@ fn parse_and_output_man_pages(
...
+        $(
+        impl From<$typ> for ManType {
+            fn from(t: $typ) -> Self {
+                ManType::$typ(t)
+            }
+        }
+        )*
+
+        #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+        enum ManType {
```
PureW: Tiny nit: Could the `enum ManType` be the first thing declared in the macro? Unless there's a reason for putting the `impl From ...` first?

src/main.rs
```diff
@@ -1187,31 +1143,102 @@ fn parse_and_output_man_pages(
...
+            $($typ($typ),)*
+        }
+
+        impl ManType {
+            const ALL: &'static [ManType] = &[$(ManType::$typ($typ),)*];
```
PureW: Just to verify that I read this correctly: This creates a single static array of all the available ManType's, on the `ManType` struct/"namespace"?

scooter-dangle: Yeap. Replaces the previous vector explicitly constructed by `parsers_to_try`.

PR #30, opened 2019-04-15
feat: add file_is_overwritable function and tests

AlanSwenson:


src/main.rs
```diff
@@ -908,31 +908,67 @@ impl TypeDeroff {
...
+    use std::io::BufRead;
+    use std::io::BufReader;
+    let display = path.display();
+    let f = match File::open(path) {
+        Err(why) => panic!("couldn't open {}: {}", display, why.description()),
+        Ok(file) => file,
+    };
+    let file = BufReader::new(&f);
+    let mut result = false;
+    for line in file.lines() {
```
scooter-dangle: Note, we'll want to use something like `split` here. Otherwise we'll encounter an otherwise resolvable panic if we read non-UTF8 bytes. Would also likely be worth it to pull in the `bstr` crate as a dependency so that we can use methods like [`trim`](https://docs.rs/bstr/0.1.2/bstr/struct.BStr.html#method.trim) on not-necessarily-UTF8 slices of bytes

Could iteraterify the loop as follows:
```rust
    Ok(file
        .split(b'\n')
        .map(|line| {
            // Okay to panic via `expect` here since we've already verified
            // that we can open the file for reading.
            bstr::B(&line.expect(&format!("I/O error encountered reading {}", display)))
                .trim()
                .to_owned()
        })
        .filter(|line| !line.is_empty())
        .take_while(|line| line.starts_with("#"))
        .any(|line| line.contains("Autogenerated")))
    // Note no need for separate `result`...we directly return the result of the iterator
```

src/main.rs
```diff
@@ -1426,3 +1462,28 @@ fn main() -> Result<(), String> {
...
+        };
+        Ok(())
+    }
+
+    pub fn remove_test_file(path: &Path) -> std::io::Result<()> {
+        use std::fs;
+        fs::remove_file(path)?;
+        Ok(())
+    }
+
```
scooter-dangle: Nit: Unnecessary newline here.

src/main.rs
```diff
@@ -908,31 +908,67 @@ impl TypeDeroff {
...
+    assert!(!file_is_overwritable(bad_path));
 
+    // Tear down
+    tests::remove_test_file(good_path);
+    tests::remove_test_file(bad_path);
+}
+
+// Return whether the file at the given path is overwritable
+// Raises IOError if it cannot be opened
 fn file_is_overwritable(path: &Path) -> bool {
```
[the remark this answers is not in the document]
PureW: Yep, that seems more rustic compared to raising a panic.

src/main.rs
PureW: > // Okay to panic via `expect` here since we've already verified
> // that we can open the file for reading.

Why is the panic okay again?

scooter-dangle: I could be wrong, but my interpretation was that the primary thing we'd need to guard against is a failure when initially opening the file and that subsequent IO errors would be unexpected enough that we don't want to build recovery logic around them. From the [documentation for split](https://doc.rust-lang.org/std/io/trait.BufRead.html#method.split):

> This function will yield errors whenever `read_until` would have also yielded an error.

Thoughts?

PureW: Gotcha, yea that makes sense.

PR #32, opened 2019-05-10
feat: App::built_command

scooter-dangle:
Who needs tests????

- - -

~Need to add tests, but will take some time to get around to that. Might still be productive to review in the intervening time.~

Added tests!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,3 +10,6 @@ itertools = "*"
 regex = "*"
 lazy_static = "*"
 bstr = "*"
+
+[dev-dependencies]
+pretty_assertions = "*"
--- a/src/main.rs
+++ b/src/main.rs
@@ -304,14 +304,14 @@ fn test_fish_options() {
 /// # Panics
 /// If `max_length` is zero.
 fn char_truncate_string(string: &str, max_length: usize, truncator: char) -> Cow<str> {
-    let mut chars = string
+    let mut char_indices = string
         .char_indices()
         .skip(max_length.checked_sub(1).unwrap())
         .map(|(idx, _)| idx);
 
-    let penultimate = chars.next();
+    let penultimate = char_indices.next();
 
-    if chars.next().is_some() {
+    if char_indices.next().is_some() {
         // Okay to unwrap since the item element _following_ `penultimate` is
         // `Some`.
         format!("{}{}", &string[0..penultimate.unwrap()], truncator).into()
--- a/src/main.rs
+++ b/src/main.rs
@@ -263,38 +263,42 @@ fn fish_options(options: &str, existing_options: &mut HashSet<String>) -> Vec<St
 
 #[test]
 fn test_fish_options() {
-    let expected_output: Vec<String> = vec!["-s 'f'".into(), "-l 'force'".into()];
-    let options = "-f, --force[=false]";
-    let mut existing_options: HashSet<String> = Default::default();
-    assert_eq!(
-        fish_options(options, &mut existing_options),
-        expected_output
-    );
-    assert_eq!(
-        expected_output.iter().cloned().collect::<HashSet<_>>(),
-        existing_options
-    );
+    {
+        let expected_output: Vec<String> = vec!["-s 'f'".into(), "-l 'force'".into()];
+        let options = "-f, --force[=false]";
+        let mut existing_options: HashSet<String> = Default::default();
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -216,93 +224,111 @@ fn complete_command(
...
-//
-//     output_complete_command(escaped_cmd, fish_options, truncated_description, built_command_output)
+const MAX_DESCRIPTION_WIDTH: usize = 78;
+const TRUNCATION_SUFFIX: &'static str = "…";
 
-// TODO args / arg types?
-fn built_command() {
-    unimplemented!()
+fn last_char_index(string: &str) -> Option<usize> {
+    string.rfind(|_| true)
```
dlrobertson: Is there a reason not to use `str::len`?

scooter-dangle: The last byte index might not not be on a UTF-8 character boundary.

dlrobertson: Ah, good to know.

scooter-dangle: Ended up not needing this. Misinterpreted the original Python's description truncation code.

Cargo.toml
```diff
@@ -10,3 +10,6 @@ itertools = "*"
 regex = "*"
 lazy_static = "*"
 bstr = "*"
+
+[dev-dependencies]
+pretty_assertions = "*"
```
justinaustin: unrelated to the PR, but will this eventually end up on crates.io? And if so do they allow wildcard dependency versions?

scooter-dangle: Hmm. Good question. Initially we'll probably only be distributing a binary, which I don't have experience with on crates.io. Not sure if it gives allowances to binaries on the manifest if they also contain a lockfile.

PureW: Assuming the library uses semantic versioning, I usually wildcard the patch-version, and lock the major and minor version. In this case, it's a dev-dependency for the unittests which decreases the risk of a complete wildcard.

Important in this context is whether you check in the `Cargo.toml` file (recommended for binaries) or not (recommended for libraries). If you don't check in the lock-file, you have a higher risk of build-failures due to changes in external dependencies.

src/main.rs
```diff
@@ -194,30 +209,213 @@ fn test_lossy_unicode() {
...
+            (if option.len() == 2 { "s" } else { "o" }, 1)
+        } else {
+            continue;
+        };
+
+        let option = format!(
+            "-{} {}",
+            fish_opt,
+            // Direct indexing of `option` won't panic due to how `num_dashes`
+            // is calculated.
```
PureW: You should add `(I promise)`.

src/main.rs
```diff
@@ -194,30 +209,213 @@ fn test_lossy_unicode() {
...
+            out.push(option);
+        }
+    }
+
+    out
+}
+
+#[test]
+fn test_fish_options() {
+    let output: Vec<String> = vec!["-s 'f'".into(), "-l 'force'".into()];
```
PureW: Tiny nit, it's usually helpful to casual readers to name the _true_ answer to something more descriptive like `expected_output` or similar.

src/main.rs
```diff
@@ -194,30 +209,213 @@ fn test_lossy_unicode() {
...
+    let mut existing_options: HashSet<String> = Default::default();
+    assert_eq!(fish_options(options, &mut existing_options), output);
+    assert_eq!(
+        output.iter().cloned().collect::<HashSet<_>>(),
+        existing_options
+    );
+
+    use std::iter::once;
+
+    let output: Vec<String> = vec!["-l 'force'".into()];
```
PureW: Another tiny tiny nit: Perhaps enclose line 277-298 and 266-273 in braces to lower risk of stray variables doing naughty things?

src/main.rs
```diff
@@ -194,30 +209,213 @@ fn test_lossy_unicode() {
...
+            .chain(once("-l 'something'".to_string()))
+            .collect::<HashSet<_>>(),
+        existing_options,
+    );
+}
+
+/// # Panics
+/// If `max_length` is zero.
+fn char_truncate_string(string: &str, max_length: usize, truncator: char) -> Cow<str> {
+    let mut chars = string
```
PureW: `s/chars/char_indices/`?

scooter-dangle:
Thanks for the review, @justinaustin! Good to go, @PureW!

PR #33, opened 2019-05-10
Adds skip_char, skip_leading_whitespace, str_at, is_white to deroff.rs

PureW:


src/deroff.rs
```diff
@@ -273,6 +273,51 @@ impl Deroffer {
...
+        let amount = amount.unwrap_or(1);
+        s.get(amount..).unwrap_or("")
+    }
+
+    //     def skip_leading_whitespace(self):
+    //         self.s = self.s.lstrip()
+    fn skip_leading_whitespace<'a>(&self, s: &'a str) -> &'a str {
+        s.trim_start()
+    }
+    //     def str_at(self, idx):
```
dlrobertson: nit! newline before comment

src/deroff.rs
```diff
@@ -273,6 +273,51 @@ impl Deroffer {
...
+        s.get(idx..idx + 1).unwrap_or("")
+    }
+
+    //     def is_white(self, idx):
+    //         # Note this returns false for empty strings (idx >= len(self.s))
+    //         return self.s[idx:idx+1].isspace()
+    fn is_white<'a>(s: &'a str, idx: usize) -> bool {
+        match Self::str_at(s, idx) {
+            "" => false,
+            c => c.chars().all(|c| c.is_ascii_whitespace()),
```
dlrobertson: Do we want to account for unicode and use `is_whitespace()`

PureW: Good question, that also affects https://github.com/rust-dc/fish-manpage-completions/pull/33/files#diff-5b0804df3203e0b4fc4f245618d61d53R291

Have we decided yet if we'll parse everything, potentially lossy to utf?

src/deroff.rs
PureW: I'd hoped `cargo fmt` would do this, I don't think too much about formatting anymore :)

src/deroff.rs
```diff
@@ -273,6 +273,53 @@ impl Deroffer {
             _ => return None,
         })
     }
+    //     def skip_char(self, amt=1):
```
scooter-dangle: Can you remove the commented Python for the functions you're implementing?

PR #34, opened 2019-05-10
Type1 type signature cleanup

justinaustin:
Unsure if we should prefer `&str` vs `String` for `options_section`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -448,7 +448,6 @@ impl ManParser for Type1 {
 //         return True
 
 impl Type1 {
-    // TODO Type signature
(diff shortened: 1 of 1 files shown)


scooter-dangle:
> Unsure if we should prefer `&str` vs `String` for `options_section`

I think `&str` make sense for now.

src/main.rs
```diff
@@ -449,7 +449,7 @@ impl ManParser for Type1 {
 
 impl Type1 {
     // TODO Type signature
```
scooter-dangle: Mind removing the type signature TODOs now that you've fixed these?

justinaustin: fixed!

PR #37, opened 2019-06-14
Add various regular expressions in deroffer (g_re_*)

necaris:


src/deroff.rs
```diff
@@ -1,10 +1,40 @@
...
 // """ Deroff.py, ported to Python from the venerable deroff.c """
+extern crate regex;
+use regex::Regex;
 
 // class Deroffer:
-struct Deroffer {}
+struct Deroffer {
+    g_re_word: &'static Regex,
+    g_re_number: &'static Regex,
+    g_re_esc_char: &'static Regex,
```
scooter-dangle: Doesn't look like `g_re_esc_char` is used anywhere. Can remove?

src/deroff.rs
```diff
@@ -1,10 +1,40 @@
 /// A translation of https://github.com/fish-shell/fish-shell/blob/e7bfd1d71ca54df726a4f1ea14bd6b0957b75752/share/tools/deroff.py
 // """ Deroff.py, ported to Python from the venerable deroff.c """
+extern crate regex;
+use regex::Regex;
 
 // class Deroffer:
-struct Deroffer {}
+struct Deroffer {
```
scooter-dangle: Instead of a struct for these regexes, could we just make all of them global statics?

src/deroff.rs
```diff
@@ -1,10 +1,40 @@
...
+                r##"(?x)([a-zA-Z_]) |  # word
+                    ([+-]?\d+)      |  # number
+                    \\                 # backslash for escape seqs
+                    "##
+            ),
+            // sequence of not backslash or whitespace
+            g_re_not_backslash_or_whitespace: crate::regex!(r##"[^ \t\n\r\f\v\\]+"##),
+            g_re_newline_collapse: crate::regex!(r##"\n{3,}"##),
+            g_re_font: crate::regex!(
+                r##"\\f(         # Starts with backslash f
```
scooter-dangle: Regex should start with `(?x)`.

src/util.rs
```diff
@@ -0,0 +1,11 @@
+extern crate regex;
```
scooter-dangle: Don't need to declare crate when using 2018 edition.

src/deroff.rs
```diff
@@ -1,10 +1,40 @@
 /// A translation of https://github.com/fish-shell/fish-shell/blob/e7bfd1d71ca54df726a4f1ea14bd6b0957b75752/share/tools/deroff.py
 // """ Deroff.py, ported to Python from the venerable deroff.c """
+extern crate regex;
```
scooter-dangle: Don't need to declare crate when using 2018 edition.

src/deroff.rs
necaris: Done in 5e03a24 :-)

src/deroff.rs
necaris: Done in 5e03a24 :-)

src/util.rs
necaris: Done in 5e03a24 :-)

src/deroff.rs
necaris: Done in 5e03a24 :-)

src/deroff.rs
necaris: In the interest of maintaining parity with the Python implementation, perhaps we could leave as-is right now? I have a sense that some of the regexes may turn out not to be necessary down the road...

PR #38, opened 2019-06-14
Feature/cleanup autogenerated file

AlanSwenson:


src/main.rs
```diff
@@ -990,14 +990,51 @@ fn file_is_overwritable(path: &Path) -> Result<bool, String> {
...
-// # Delete the file if it is autogenerated
-// def cleanup_autogenerated_file(path):
-//     try:
-//         if file_is_overwritable(path):
-//             os.remove(path)
-//     except (OSError, IOError):
-//         pass
+    // Setup file paths
+    // Good File Paths
+    let good_string = String::from("good.txt");
```
scooter-dangle: Hmm. We'll need to rethink this test strategy. Since the tests are run in parallel, I get intermittent failures due multiple tests adding/removing files with the same path concurrently.

scooter-dangle:
> @scooter-dangle @dlrobertson what do I do with the errors here? In original python version we except them then pass..

Sorry I didn't see this before!

For other peeps' reference, @AlanSwenson and I talked on Thursday and went with mimicking the Python behavior by explicitly swallowing the error (via the `let _ = some_result_expression` pattern).

scooter-dangle:
> > For other peeps' reference, @AlanSwenson and I talked on Thursday and went with mimicking the Python behavior by explicitly swallowing the error (via the `let _ = some_result_expression` pattern).
> 
> Would we want to log it to `stderr` to get visibility into failures?

Ooooh. Yeah, that could be really useful. Does that make sense, @AlanSwenson? Maybe something like
```rust
fn cleanup_autogenerated_file(path: &Path) {
    use std::error::Error;

    if file_is_overwritable(path) == Ok(true) {
        // Original proceeds while ignoring errors
        if let Err(err) = std::fs::remove_file(path) {
            eprintln!(
                "Error in cleaning up auto-generated file ({}): {}",
                path.to_string_lossy(),
                err.description(),
            );
        }
    }
}
```

(the `use std::error::Error` is necessary so that we can call `Error::description` on `err`)

scooter-dangle:
For the parallel tests issue, I think it would be good to ensure we don't have to move to single-threaded tests. We won't see any difference in runtime now, but we probably will over time as more tests are added, and it will be more difficult at that point to possibly fix a number of tests that assume they can run on their own rather than avoid that now.

I played with something like the following:

```diff
diff --git a/src/main.rs b/src/main.rs
index 4ef8811..f256f60 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -920,14 +920,15 @@ impl TypeDeroff {
 #[test]
 fn test_file_is_overwritable() {
     use std::path::Path;
+    use tests::FileKind;
 
     // Setup file paths
     // Good File Paths
-    let good_string = String::from("good.txt");
+    let good_string = String::from("test_file_is_overwritable__good.txt");
     let good_path = Path::new(&good_string);
 
     // Bad File paths
-    let bad_string = String::from("bad.txt");
+    let bad_string = String::from("test_file_is_overwritable__bad.txt");
     let bad_path = Path::new(&bad_string);
 
     // Remove any leftover files if the already
@@ -936,14 +937,14 @@ fn test_file_is_overwritable() {
     tests::remove_test_file(bad_path);
 
     // Create good test file
-    tests::create_test_file(good_path);
+    tests::create_test_file(good_path, FileKind::Good);
 
     // Test for IO Error when File doesn't exist
     let result = file_is_overwritable(bad_path);
     assert!(result.is_err());
 
     // Create bad test file
-    tests::create_test_file(bad_path);
+    tests::create_test_file(bad_path, FileKind::Bad);
 
     // Tests
     let result = file_is_overwritable(good_path);
@@ -993,14 +994,15 @@ fn file_is_overwritable(path: &Path) -> Result<bool, String> {
 #[test]
 fn test_cleanup_autogenerated_file() {
     use std::path::Path;
+    use tests::FileKind;
 
     // Setup file paths
     // Good File Paths
-    let good_string = String::from("good.txt");
+    let good_string = String::from("test_cleanup_autogenerated_file__good.txt");
     let good_path = Path::new(&good_string);
 
     // Bad File paths
-    let bad_string = String::from("bad.txt");
+    let bad_string = String::from("test_cleanup_autogenerated_file__bad.txt");
     let bad_path = Path::new(&bad_string);
 
     // Remove any leftover files if the already
@@ -1009,14 +1011,14 @@ fn test_cleanup_autogenerated_file() {
     tests::remove_test_file(bad_path);
 
     // Create good test file
-    tests::create_test_file(good_path);
+    tests::create_test_file(good_path, FileKind::Good);
 
     // Test for IO Error when File doesn't exist
     //let result = cleanup_autogenerated_file(bad_path);
     // assert!(result.is_err());
 
     // Create bad test file
-    tests::create_test_file(bad_path);
+    tests::create_test_file(bad_path, FileKind::Bad);
 
     // Tests
     cleanup_autogenerated_file(good_path);
@@ -1513,16 +1515,29 @@ fn main() -> Result<(), String> {
 mod tests {
     use super::*;
 
-    pub fn create_test_file(path: &Path) -> std::io::Result<()> {
+    #[derive(Copy, Clone, Debug)]
+    pub enum FileKind {
+        Good,
+        Bad,
+    }
+
+    impl FileKind {
+        fn content(self) -> &'static [u8] {
+            match self {
+                FileKind::Good => {
+                    b"    \n#Hello, world!\
+                    \n#Hello, world! Autogenerated      "
+                }
+                FileKind::Bad => b"    Autogenerated     ",
+            }
+        }
+    }
+
+    pub fn create_test_file(path: &Path, kind: FileKind) -> std::io::Result<()> {
         use std::fs::File;
         use std::io::prelude::*;
         let mut file = File::create(path)?;
-        if path.to_str() == Some("bad.txt") {
-            file.write_all(b"    Autogenerated     ")?;
-        } else {
-            file.write_all(b"    \n#Hello, world!\n")?;
-            file.write_all(b"#Hello, world! Autogenerated      ")?;
-        };
+        file.write_all(kind.content())?;
         Ok(())
     }
 
```

Thoughts?

AlanSwenson:
@scooter-dangle great idea for the fileKind. Changes made and submitted.

src/main.rs
```diff
@@ -979,6 +980,27 @@ fn file_is_overwritable(path: &Path) -> Result<bool, String> {
         .take_while(|line| line.starts_with("#"))
         .any(|line| line.contains("Autogenerated")))
 }
+#[test]
+fn test_cleanup_autogenerated_completions_in_directory() {
+    let bad_string = String::from(".");
+    let bad_path = Path::new(&bad_string);
+    // need to create a directory path with files for testing
+    //cleanup_autogenerated_completions_in_directory(bad_path);
```
scooter-dangle: Maybe we can discuss strategies for testing+cleanup at the meetup? Could also move `cleanup_autogenerated_completions_in_directory` to a new PR since I think the rest of this branch looks good to merge.

AlanSwenson: yeah, this was meant for a new PR....oops

AlanSwenson: removed

PR #39, opened 2019-06-14
Some cleanup and Deroffer properties

PureW:
Some early cleanup and structure to the Deroffer struct.

### DIFF
--- a/src/deroff.rs
+++ b/src/deroff.rs
@@ -541,7 +541,7 @@ fn test_is_white() {
 //     def str_eq(offset, other, len):
 //         return self.s[offset:offset+len] == other[:len]
 
-//     def prch(self, idx):
+//     def not_whitespace(self, idx):
 //         # Note that this return False for the empty string (idx >= len(self.s))
 //         ch = self.s[idx:idx+1]
 //         return ch not in ' \t\n'
@@ -555,14 +555,14 @@ fn test_is_white() {
 //     def font2(self):
 //         if self.s[0:2] == '\\f':
 //             c = self.str_at(2)
-//             if c == '(' and self.prch(3) and self.prch(4):
+//             if c == '(' and self.not_whitespace(3) and self.not_whitespace(4):
 //                 self.skip_char(5)
 //                 return True
 //             elif c == '[':
 //                 self.skip_char(2)
-//                 while self.prch(0) and self.str_at(0) != ']': self.skip_char()
+//                 while self.not_whitespace(0) and self.str_at(0) != ']': self.skip_char()
 //                 if self.str_at(0) == ']': self.skip_char()
-//             elif self.prch(2):
+//             elif self.not_whitespace(2):
 //                 self.skip_char(3)
 //                 return True
 //         return False
@@ -592,22 +592,22 @@ fn test_is_white() {
 //             if self.s[3:5] == 'dy':
 //                 self.skip_char(5)
 //                 return True
-//             elif self.str_at(2) == '(' and self.prch(3) and self.prch(4):
+//             elif self.str_at(2) == '(' and self.not_whitespace(3) and self.not_whitespace(4):
 //                 self.skip_char(5)
 //                 return True
-//             elif self.str_at(2) == '[' and self.prch(3):
+//             elif self.str_at(2) == '[' and self.not_whitespace(3):
 //                 self.skip_char(3)
 //                 while self.str_at(0) and self.str_at(0) != ']':
 //                     self.skip_char()
 //                 return True
-//             elif self.prch(2):
+//             elif self.not_whitespace(2):
 //                 self.skip_char(3)
 //                 return True
 //         elif s0s1 == '\\*':
-//             if self.str_at(2) == '(' and self.prch(3) and self.prch(4):
+//             if self.str_at(2) == '(' and self.not_whitespace(3) and self.not_whitespace(4):
 //                 reg = self.s[3:5]
 //                 self.skip_char(5)
-//             elif self.str_at(2) == '[' and self.prch(3):
+//             elif self.str_at(2) == '[' and self.not_whitespace(3):
 //                 self.skip_char(3)
 //                 while self.str_at(0) and self.str_at(0) != ']':
 //                     reg = reg + self.str_at(0)
@@ -616,7 +616,7 @@ fn test_is_white() {
 //                     self.skip_char()
 //                 else:
 //                     return False
-//             elif self.prch(2):
+//             elif self.not_whitespace(2):
 //                 reg = self.str_at(2)
 //                 self.skip_char(3)
 //             else:
@@ -639,7 +639,7 @@ fn test_is_white() {
 
 //     def spec(self):
 //         self.specletter = False
-//         if self.s[0:2] == '\\(' and self.prch(2) and self.prch(3):
+//         if self.s[0:2] == '\\(' and self.not_whitespace(2) and self.not_whitespace(3):
 //             key = self.s[2:4]
 //             if key in Deroffer.g_specs_specletter:
 //                 self.condputs(Deroffer.g_specs_specletter[key])
--- a/src/deroff.rs
+++ b/src/deroff.rs
@@ -8,9 +8,14 @@ use std::collections::HashMap;
 type TODO_TYPE = u8;
 type TODO_NUMBER_TYPE = i8;
 
-const OPTIONS: u8 = 0;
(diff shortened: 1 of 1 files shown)


src/deroff.rs
```diff
@@ -1,8 +1,40 @@
 /// A translation of https://github.com/fish-shell/fish-shell/blob/e7bfd1d71ca54df726a4f1ea14bd6b0957b75752/share/tools/deroff.py
 // """ Deroff.py, ported to Python from the venerable deroff.c """
+use std::collections::HashMap;
+
+type TODO_TYPE = u8;
+type TODO_NUMBER_TYPE = i8;
+
+const OPTIONS: u8 = 0;
```
scooter-dangle: Looks like these three constants are exclusively used by the `tblstate` field (and _it_ only uses these three constants). Maybe we could convert them into
```rust
enum TblState {
    Options,
    Format,
    Data,
}
```
and change the `tblstate` field's type to `TblState`.

src/deroff.rs
```diff
@@ -1,8 +1,40 @@
...
-struct Deroffer {}
+struct Deroffer {
+    reg_table: HashMap<TODO_TYPE, TODO_TYPE>,
+    tr_from: String,
+    tr_to: String,
+    tr: String,
+    nls: TODO_NUMBER_TYPE,
+    specletter: bool,
+    refer: bool,
+    r#macro: TODO_NUMBER_TYPE,
```
scooter-dangle: I'm not positive, but it looks like the `macro` field might only ever be `0` or `1`. Is it possible it's just a boolean? (Can leave as is if we're not sure.)

src/deroff.rs
```diff
@@ -1,8 +1,40 @@
...
+const DATA: u8 = 2;
 
 // class Deroffer:
-struct Deroffer {}
+struct Deroffer {
+    reg_table: HashMap<TODO_TYPE, TODO_TYPE>,
+    tr_from: String,
+    tr_to: String,
+    tr: String,
+    nls: TODO_NUMBER_TYPE,
```
scooter-dangle: Unless there's some indirect field access, it doesn't look like `nls` is ever used aside from initializing it to `2`. Should be able to remove.

src/deroff.rs
```diff
@@ -1,8 +1,40 @@
...
+    inlist: bool,
+    inheader: bool,
+    pic: bool,
+    tbl: bool,
+    tblstate: TODO_NUMBER_TYPE,
+    tblTab: String,
+    eqn: bool,
+    skipheaders: bool,
+    skiplists: bool,
+    ignore_sonx: bool,
```
scooter-dangle: Never used. Can remove.

src/deroff.rs
```diff
@@ -1,8 +1,40 @@
...
+    nobody: bool,
+    inlist: bool,
+    inheader: bool,
+    pic: bool,
+    tbl: bool,
+    tblstate: TODO_NUMBER_TYPE,
+    tblTab: String,
+    eqn: bool,
+    skipheaders: bool,
+    skiplists: bool,
```
scooter-dangle: Looks like `skiplists` is always `False` in the original code. Can make it a global constant? (Maybe it's a useful toggle somehow in debugging?)

src/deroff.rs
```diff
@@ -296,6 +328,27 @@ impl Deroffer {
             c => c.chars().all(|c| c.is_whitespace()),
         }
     }
+
+    fn not_whitespace(s: &str, idx: usize) -> bool {
+        //         # Note that this return False for the empty string (idx >= len(self.s))
+        //         ch = self.s[idx:idx+1]
+        //         return ch not in ' \t\n'
+        let ch = s
```
scooter-dangle: Could maybe succinctify the function body a little:
```rust
        s.get(idx..(idx + 1))
            .map(|string| string == " " || string == "\t" || string == "\n")
            .unwrap_or_default()
```

src/deroff.rs
```diff
@@ -1,8 +1,40 @@
 /// A translation of https://github.com/fish-shell/fish-shell/blob/e7bfd1d71ca54df726a4f1ea14bd6b0957b75752/share/tools/deroff.py
 // """ Deroff.py, ported to Python from the venerable deroff.c """
+use std::collections::HashMap;
+
+type TODO_TYPE = u8;
```
pickfire: Types should be `UpperCamelCase` as per https://rust-lang-nursery.github.io/api-guidelines/naming.html

src/deroff.rs
pickfire: My suggestion
```rust
        s.get(idx..(idx + 1))
            .map(|s| s.chars().next())
            .map_or(false, |ch| ch == ' ' || ch == '\t' || ch == '\n')
```

src/deroff.rs
```diff
@@ -296,6 +328,27 @@ impl Deroffer {
             c => c.chars().all(|c| c.is_whitespace()),
         }
     }
+
+    fn not_whitespace(s: &str, idx: usize) -> bool {
```
pickfire: From what I think, we should check not whitespace for a range, not an index as the below calls it with `and`.
```rust
//          if self.s[0:2] == '\\(' and self.not_whitespace(2) and self.not_whitespace(3):
```
It could have just been
```rust
//          if self.s[0:2] == '\\(' and self.not_whitespace(2..4):
```
As well, why not just use `is_ascii_whitespace()`?

src/deroff.rs
scooter-dangle: Thanks for commenting!

> My suggestion
> 
> ```rust
>         s.get(idx..(idx + 1))
>             .map(|s| s.chars().next())
> ```

In that line there you'd want to use `and_then` instead of `map`, since `next` is going to return and `Option<char>` which will be wrapped in `Some` by `map` to be `Option<Option<char>>`. Could also `unwrap` since the indexing restrictions in `get` ensure that we _have_ to have a single `char` in `s`. I also rely on that in my version to allow `"...".contains(...)` to work—it only does what we want because if `get(idx..(idx + 1))` returns a `Some`, it must be a valid single-byte (and so also single-char) UTF-8 string.

> ```rust
>             .map_or(false, |ch| ch == ' ' || ch == '\t' || ch == '\n')
> ```

I'm totally cheating in my suggestion, but I don't know whether there's a performance penalty for that cheat. It just lets me avoid the longer combination of individual equality checks, but isn't necessarily performant.

If we didn't want to cheat by using `str::contains`, I'd likely go with

```rust
        s.get(idx..(idx + 1))
            .map(|string| string == " " || string == "\t" || string == "\n")
            .unwrap_or_default()
```

Up to @PureW whether he wants to modify his original.

src/deroff.rs
scooter-dangle: In general, I agree. I think as a placeholder/todo type declaration the screaming snake case is probably fine.

src/deroff.rs
scooter-dangle: > From what I think, we should check not whitespace for a range, not an index as the below calls it with `and`.
> 
> ```rust
> //          if self.s[0:2] == '\\(' and self.not_whitespace(2) and self.not_whitespace(3):
> ```
> 
> It could have just been
> 
> ```rust
> //          if self.s[0:2] == '\\(' and self.not_whitespace(2..4):
> ```

That's a good point. I hadn't looked closely at where `not_whitespace` was being used, but that makes sense. The one wrinkle is that the original function explicitly returns `false` if the target string is empty, which would complicate a range/iterator implementation. It would need to do a length check in addition to the character check to ensure that the actual substring length is the same as the range length. Might be something to revisit after we finish the initial port.

> As well, why not just use `is_ascii_whitespace()`?

Good question. For the initial port we're _largely_ replicating the literal behavior of the original Python source, which only checks for `' '`, `'\t'` and `'\n'`, whereas the `is_ascii_whitespace` methods in Rust return `true` for a number of other characters. I haven't looked closely enough at how the function is being used to know whether we could get away with `is_ascii_whitespace`. It's something we might be able to do after we have our port working end-to-end.

src/deroff.rs
PureW: I guess so, the `numreq` function sure does look weird.

src/deroff.rs
PureW: Nice discussion! I went with the first suggestion as it seemed the easiest to read after the fact.

src/deroff.rs
PureW: Added a TODO to revisit ascii whitespace after MVP.

PR #40, opened 2019-06-14
Implement g_macro_dict as g_macro_dispatch

necaris:


src/deroff.rs
```diff
@@ -296,6 +326,152 @@ impl Deroffer {
...
+            "so" => self.macro_so_nx(),
+            "nx" => self.macro_so_nx(),
+            "tr" => self.macro_tr(),
+            "sp" => self.macro_s(),
+            _ => self.macro_other(),
+        }
+    }
+
+    fn macro_sh(&mut self) -> bool {
+        false
```
scooter-dangle: Might be more accurate for all of the default function bodies to be `unimplemented!()` rather than `false`.

src/deroff.rs
```diff
@@ -296,6 +326,152 @@ impl Deroffer {
             c => c.chars().all(|c| c.is_whitespace()),
         }
     }
+
+    // Replaces the g_macro_dict lookup in the Python code
+    fn g_macro_dispatch<'a>(&mut self, s: &'a str) -> bool {
```
scooter-dangle: Lifetime annotation isn't required here.
```suggestion
    fn g_macro_dispatch(&mut self, s: &str) -> bool {
```

scooter-dangle:
Looks good overall. The comments on #37 also apply, but don't need to be resolved here if we merge that first and rebase this one.

PR #43, opened 2019-06-19
solve parallel test problem by using 1 thread

AlanSwenson:
Possible solution for the testing problem brought up in #38


scooter-dangle:
See https://github.com/rust-dc/fish-manpage-completions/pull/38#issuecomment-503783893

scooter-dangle:
Modifications were made in #38 to make it more straightforward for each test to use it's own test files, which resolves the concurrency issue.

PR #45, opened 2019-07-12
Add deroff_files function

justinaustin:
I didn't realize that the `deroff` and `flush_output` methods haven't been ported over yet. Unsure if it would be better to hold off merging this until those have been completed.

Original python code:
```python
 def deroff_files(files):
     for arg in files:
         sys.stderr.write(arg + '\n')
         if arg.endswith('.gz'):
             f = gzip.open(arg, 'r')
             str = f.read()
             if IS_PY3: str = str.decode('latin-1')
         else:
             f = open(arg, 'r')
             str = f.read()
         d = Deroffer()
         d.deroff(str)
         d.flush_output(sys.stdout)
         f.close()
```


src/deroff.rs
```diff
@@ -530,6 +533,27 @@ impl Deroffer {
     }
 }
 
+fn deroff_files(files: &[String]) -> std::io::Result<()> {
+    for arg in files {
+        eprintln!("{}", arg);
```
PureW: Perhaps add some informational string before `{}` so it becomes easier to backtrack later from program output to location in source code?

src/deroff.rs
```diff
@@ -530,6 +533,27 @@ impl Deroffer {
...
+        if arg.ends_with(".gz") {
+            let mut decoder = Decoder::new(file).unwrap();
+            decoder.read_to_string(&mut string);
+        } else {
+            file.read_to_string(&mut string)?;
+        }
+        let d = Deroffer::new();
+        println!("string: {}", string);
+
+        // d.deroff(string);
```
PureW: Perhaps a `// TODO Enable when implemented` comment?

scooter-dangle:
Could you remove the corresponding commented out Python code as part of this branch?

scooter-dangle:
> I didn't realize that the `deroff` and `flush_output` methods haven't been ported over yet. Unsure if it would be better to hold off merging this until those have been completed.

Nah. No need to hold off. If this looks equivalent to the original we'll be able to go ahead and merge.

src/deroff.rs
scooter-dangle: For type-checking could maybe at least add in the function type signatures but leave the body's `unimplemented!()`?

src/deroff.rs
```diff
@@ -530,6 +533,27 @@ impl Deroffer {
...
 }
 
+fn deroff_files(files: &[String]) -> std::io::Result<()> {
+    for arg in files {
+        eprintln!("processing deroff file: {}", arg);
+
+        let mut file = File::open(arg)?;
+        let mut string = String::new();
+        if arg.ends_with(".gz") {
+            let mut decoder = Decoder::new(file).unwrap();
```
scooter-dangle: What are the conditions for which `Decoder::new` would return an `Err`? (When would the `unwrap` panic?)

src/deroff.rs
justinaustin: added!

src/deroff.rs
justinaustin: added!

src/deroff.rs
justinaustin: unsure...the documentation for libflate seems to be lacking :( https://docs.rs/libflate/0.1.25/libflate/gzip/struct.Decoder.html

scooter-dangle: Blerg. Yeah. Looks like it's IO errors or the file not being compressed via `DEFLATE(8)`. Probly okay to leave for now. Don't know how often this will happen.

PR #46, opened 2019-07-12
Add deroff fn letter

pmatseykanets:


pmatseykanets:
@scooter-dangle @PureW please take a look. Thanks!

src/deroff.rs
```diff
@@ -376,6 +376,13 @@ impl Deroffer {
         }
     }
 
+    fn letter<'a>(s: &'a str, idx: usize) -> bool {
+        match Self::str_at(s, idx) {
+            "" => false,
+            c => c.chars().all(|c| c.is_alphabetic() || c == '_'), // underscore is used in C identifiers
```
pickfire:
```suggestion
        s.get(idx..idx+1)
            .map(s => s.chars.next())
            .map_or(false, c => c.is_ascii_alphabetic() || c == '_') // underscore is used in C identifiers
```

I would not use `str_at` because of the unnecessary `unwrap_or("")` which needs to be checked again later.

As for `c.chars().all()`, we I just get a single character so no point to iterate over a single character.

I would be better to just use `is_ascii_alphabetic` to not include a whole range of other characters.

With the change, the other blocks are not needed.

scooter-dangle:
Kinda embarrassing to realize this _now_, but it doesn't look like `letter` is actually used anywhere. Can modify PR to just remove the commented-out Python source. 🤦‍♂

src/deroff.rs
scooter-dangle: > I would not use `str_at` because of the unnecessary `unwrap_or("")` which needs to be checked again later.
> 
> As for `c.chars().all()`, we I just get a single character so no point to iterate over a single character.

That's a good point on `str_at`. It might make more sense for `str_at` to return `Option<char>` (where `None` expresses the currently possible `""` return value). Would simplify things. Of course, that still might change depending on whether we stick with UTF-8 string processing rather than byte-string processing in `deroff.rs`. I might try that out after #46 and #47 are merged.

For _this_ PR (and #47) with the current `&str` return value of `str_at`, I think it's fine to use `c.chars().all(...)`. It seems idiomatic...would consider something different if benchmarking showed a different form could noticeably affect performance.

> I would be better to just use `is_ascii_alphabetic` to not include a whole range of other characters.

This might be what we want to do if we switch to byte-strings rather than strings. I _think_ the current formulation is an accurate port in the context of non-ASCII UTF-8 characters.

scooter-dangle: @pickfire, does that make sense—converting `str_at` to return `Option<char>` like

```rust
    fn str_at(string: &str, idx: usize) -> Option<char> {
        string.chars().skip(idx).next()
    }
```

?

pickfire: Yes, only if one character is desired from my point of view. I prefer clarity if of having an character (`Some('a')`) and none (`None`) instead of implicit `''` (not sure if that works).

If performance advise is needed, maybe someone like @BurntSushi could give some advise on this.

PR #47, opened 2019-07-12
Add deroff fn digit

pmatseykanets:
Original implementation

```py
def digit(self, idx):
     ch = self.str_at(idx)
     return ch.isdigit()
```

### DIFF
--- a/src/deroff.rs
+++ b/src/deroff.rs
@@ -376,6 +376,13 @@ impl Deroffer {
         }
     }
 
+    fn digit<'a>(s: &'a str, idx: usize) -> bool {
+        match Self::str_at(s, idx) {
+            "" => false,
+            c => c.chars().all(|c| c.is_digit(10)),
+        }
+    }
+
     // Replaces the g_macro_dict lookup in the Python code
     fn g_macro_dispatch(&mut self, s: &str) -> bool {
         match s {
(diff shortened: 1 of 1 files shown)


pmatseykanets:
@scooter-dangle @PureW please take a look. Thanks!

src/deroff.rs
```diff
@@ -376,6 +376,13 @@ impl Deroffer {
         }
     }
 
+    fn digit<'a>(s: &'a str, idx: usize) -> bool {
+        match Self::str_at(s, idx) {
+            "" => false,
+            c => c.chars().all(|c| c.is_digit(10)),
```
pickfire: The suggestions in https://github.com/rust-dc/fish-manpage-completions/pull/46/files#r303175909 applies here as well.

In addition, it would be good to just use `is_ascii_digit()`.

scooter-dangle: Same response as at https://github.com/rust-dc/fish-manpage-completions/pull/46/files#r304206971

Will look at converting `str_at` return value to `Option<char>` after this PR is merged.

`char::is_digit` is fine for now. Can revisit to verify behavior when we're closer to having a working Proof-of-Concept.

src/deroff.rs
```diff
@@ -376,6 +376,13 @@ impl Deroffer {
         }
     }
 
+    fn digit<'a>(s: &'a str, idx: usize) -> bool {
```
scooter-dangle: Lifetime annotation can be elided here.
```suggestion
    fn digit(s: &str, idx: usize) -> bool {
```

PR #55, opened 2019-09-13
feat: implement deroff::digit

scooter-dangle:
Subsumes #47

### DIFF
--- a/src/deroff.rs
+++ b/src/deroff.rs
@@ -392,6 +392,13 @@ impl Deroffer {
         }
     }
 
+    fn digit(s: &str, idx: usize) -> bool {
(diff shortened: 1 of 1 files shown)


src/deroff.rs
```diff
@@ -392,6 +392,13 @@ impl Deroffer {
         }
     }
 
+    fn digit(s: &str, idx: usize) -> bool {
+        match Self::str_at(s, idx) {
+            "" => false,
+            c => c.chars().all(|c| c.is_digit(10)),
+        }
```
pickfire: Why is checking `chars()` required here? Isn't `str_at` returning just one character in str format? Can we just access the first item and check if it is digit?

PR #56, opened 2019-09-13
Some macros

kbknapp:


kbknapp:
@scooter-dangle should be good now

src/deroff.rs
```diff
@@ -1299,3 +1331,42 @@ fn test_is_white() {
...
+            panic!("from and to must be the same length");
+        }
+        for (f, t) in from.chars().zip(to.chars()) {
+            self.map.insert(f.to_string(), t.to_string());
+        }
+    }
+    fn translate(&self, s: &str) -> String {
+        let mut s = s.to_owned();
+        for (k, v) in self.map.iter() {
+            s = (&*s).replace(k, v);
```
scooter-dangle: Would be good to include tests a couple of this logic to ensure it does. E.g., the following fails:
```rust
#[test]
fn test_translate_table() {
    let mut tbl = TranslateTable::new();
    tbl.maketrans("abc", "bca");
    assert_eq!(tbl.translate("abc"), "bca");
}
```
Not sure if any valid roff files ever have shared chars between the `to` and `from` strings, but wouldn't want to rely on that.

src/deroff.rs
```diff
@@ -555,6 +700,20 @@ impl Deroffer {
...
+        match s.chars().nth(0) {
+            Some('.') | Some('\'') => !request_or_macro(s),
+            Some(c) => {
+                if self.tbl {
+                    do_tbl(s)
+                } else {
+                    text(s)
+                }
+            }
+            None => unreachable!(),
```
scooter-dangle: Nit: Not locally unreachable. Maybe
```suggestion
            None => panic!("do_line` called with empty string as argument"),
```
instead?

src/deroff.rs
scooter-dangle: Obsoleted by #64

PR #58, opened 2019-10-10
Stab at esc_char and quoted_arg in deroff.

sstanfield:
Not sure about the interface I am using, went with &str in &str out (or Option wrapped to indicate nothing changed/false).  Maybe the s member of deroff should be used but some stuff seemed to be going this route.  The quoted_arg return type probably needs to change if this interface stands.


src/deroff.rs
```diff
@@ -548,6 +548,70 @@ impl Deroffer {
...
+                if let Some(ns) = self.word(s) {
+                    Some(ns)
+                } else {
+                    self.number(s)
+                }
+            }
+        } else {
+            None
+        }
+    }
```
scooter-dangle: Suggestion:

```rust
    fn esc_char(&self, s: &str) -> Option<&str> {
        s.get(0..1).and_then(|ch| {
            if ch == "\\" {
                self.esc_char_backslash(s)
            } else {
                self.word(s).or_else(|| self.number(s))
            }
        })
    }
```

src/deroff.rs
```diff
@@ -548,6 +548,70 @@ impl Deroffer {
...
+                } else {
+                    self.condputs(Deroffer::str_at(string, 0));
+                    string = self.skip_char(string, Some(1));
+                }
+            }
+            true
+        } else {
+            false
+        }
+    }
```
scooter-dangle: Some tweaks per meetup discussion:
```rust
    fn quoted_arg(&self, string: &str) -> bool {
        if Deroffer::str_at(string, 0) == "\"" {
            // We've now entered a portion of the source that should be
            // surrounded by double quotes. (We've found the first one—really
            // hoping we find its mate later).
            let mut string = self.skip_char(string, Some(1));
            while !string.is_empty() && Deroffer::str_at(string, 0) != "\"" {
                // Our string starts with _any_ char other than a double-quote
                if let Some(ns) = self.esc_char(string) {
                    // Our thing started with a backslash or was parseable as a
                    // word or a number.
                    string = ns;
                } else {
                    self.condputs(Deroffer::str_at(string, 0));
                    string = self.skip_char(string, Some(1));
                }
            }
            // We've run past the end of the string OR we've found the closing
            // double-quote to match the initial one we found at the start of
            // the function.
            true
        } else {
            // We don't start with quotes!
            false
        }
    }
```

PR #62, opened 2019-11-08
Add `util:translate` and `util:maketrans`

charliethomson:
These functions mirror their python counterparts, `translate` takes a String and a table (`HashMap<char, char>`) and remaps the String based on the table; and `maketrans` takes two equal length Strings and creates a table for use in the translate function.

These are needed a few times in the python source, so I made them (I couldn't find a crate for it)

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -16,7 +16,7 @@ use std::{
 
 /// Helper for `translate`, does the same thing as string.maketrans in python
 /// requires that f and t are the same length, or it takes the shorter one
-pub fn maketrans(f: String, t: String) -> HashMap<u8, char> {
+pub fn maketrans(f: &str, t: &str) -> HashMap<u8, char> {
     HashMap::from_iter(f.bytes().zip(t.chars()))
 }
 
@@ -43,8 +43,8 @@ fn test_maketrans() {
     expected.insert(0x63, 'f');
     assert_eq!(
         maketrans(
-            "abc".to_owned(),
-            "def".to_owned(),
+            "abc",
+            "def",
         ),
         expected
     );
@@ -61,8 +61,8 @@ fn test_translate() {
         translate(
             "Hello World!".to_owned(),
             maketrans(
-                "HlW".to_owned(),
-                "aBc".to_owned(),
+                "HlW",
+                "aBc",
             )
         ),
         expected,
--- a/src/util.rs
+++ b/src/util.rs
@@ -7,3 +7,64 @@ macro_rules! regex {
         &REGEX
     }};
 }
+
+use std::{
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -7,3 +7,64 @@ macro_rules! regex {
...
+
+use std::{
+    iter::FromIterator,
+    collections::HashMap,
+};
+
+
+/// Helper for `translate`, does the same thing as string.maketrans in python
+/// requires that f and t are the same length, or it takes the shorter one
+pub fn maketrans(f: String, t: String) -> HashMap<u8, char> {
```
pickfire: Why not consume `&str` instead of `String`?

charliethomson: Good point. `&str` makes more sense, I'll make that change

src/util.rs
```diff
@@ -7,3 +7,64 @@ macro_rules! regex {
...
+
+/// Helper for `translate`, does the same thing as string.maketrans in python
+/// requires that f and t are the same length, or it takes the shorter one
+pub fn maketrans(f: &str, t: &str) -> HashMap<u8, char> {
+    HashMap::from_iter(f.bytes().zip(t.chars()))
+}
+
+/// Does effectively the same as string.translate in python, in 
+/// conjunction with `maketrans`, basically just replaces
+pub fn translate(s: String, table: HashMap<u8, char>) -> String {
```
pickfire: I believe we can just modified the String `s` here without the need to create a new String, that saves additional allocations by just performing the operations on the existing data. I wonder if `s.bytes().map(|byte| table.get(&byte).unwrap_or(byte))` works.

charliethomson:
@pickfire
You're definitely right. I'll reformat when I get home. Thanks for the input. Also, I noticed that I never committed a change that would change the type of the output of `maketrans` and the input of `translate` to `HashMap<char, char>`, thoughts?

pickfire:
@charliethomson Looks good to me. But why not? (not tested)
```rust
     f.chars().zip(t.chars()) // or .to_iter() or .into_iter()
```

charliethomson:
@pickfire 
It's already (almost exactly) like that 
(src/util.rs:19 - 20)
```rust
pub fn maketrans(f: &str, t: &str) -> HashMap<char, char> {
    HashMap::from_iter(f.bytes().zip(t.chars()))
```

scooter-dangle:
This is great! Thanks! My bad for not spotting this earlier, but I think there is some duplication here with what's in #56. Will look at what it looks like locally merging that one before this one and vice versa.

scooter-dangle:
Going to close this PR in favor of #64, since that is a superset of what is in this PR.

PR #64, opened 2019-11-09
impl `condputs`

charliethomson:
I combined `condputs` and `condputs_tr`, because one overloads the other
later on. I also changed the type of `Deroffer:self.tr` from `String` to
`Option<HashMap<char, char>>` in order to be able to interface with
`util:translate` and `util:maketrans`. In addition, I set the type of
`Deroffer:self.output` to `String`


src/deroff.rs
```diff
@@ -538,6 +539,31 @@ impl Deroffer {
...
+            self.pic     ||
+            self.eqn     ||
+            self.refer   ||
+            self.r#macro ||
+            self.inlist  ||
+            self.inheader
+        };
+
+        if !is_special {
+            if let Some(table) = self.tr.clone() {
```
pickfire: Why is the `.clone()` needed here instead of having it below?

charliethomson: 'HashMap` isn't `Copy` which is required for `if let Some(table)` so it moves `table` out of `self.tr`, which means that `self.tr` is no longer valid, so we have to `clone` so we don't lose the translation table

(If `HashMap` was `Copy` we wouldn't need any `clones` at all)

pickfire:
Any news on this?

src/deroff.rs
scooter-dangle: To avoid having to clone, `translate` should take `&HashMap<_, _>` instead of `HashMap<_, _>`.

src/util.rs
```diff
@@ -7,3 +7,41 @@ macro_rules! regex {
         &REGEX
     }};
 }
+
+use std::{collections::HashMap, iter::FromIterator};
+
+/// Helper for `translate`, does the same thing as string.maketrans in python
+/// requires that f and t are the same length, or it takes the shorter one
+pub fn maketrans(f: &str, t: &str) -> HashMap<char, char> {
```
scooter-dangle: To match the Python semantics, `maketrans` should either panic or return an `Err` if the character lengths of the two inputs are not equal.

scooter-dangle:
Aside from the comments above, looks good overall.

It's not strictly necessary, but you might consider making the translation table a dedicated type (as @kbknapp had done in #56). If so, it might make sense to convert `maketrans` to `TranslateTable::new`, e.g.,

```rust
impl TranslateTable {
    pub fn new(f: &str, t: &str) -> Result<Self, String> {
        /* ... */
    }
    /* ... */
}
```

scooter-dangle:
Note: See additional discussion/review of some of the code in this PR in #62

scooter-dangle:
Oh, also. Could you delete the Python source for `condputs` and `condputs_tr` as part of this branch?

charliethomson:
```rust
pub struct TranslationTable {
    table: HashMap<char, char>,
} impl TranslationTable {
    pub fn new(f: &str, t: &str) -> Result<Self, &'static str> {
        if f.len() != t.len() {
            Err("Arguments passed to maketrans must have equal length")
        } else {
            Ok(TranslationTable { table: HashMap::from_iter(f.chars().zip(t.chars())) })
        }
    }

    pub fn translate(&self, s: String) -> String {
        s
            .chars()
            .map(|c| *self.table.get(&c).unwrap_or(&c))
            .collect::<String>()
    }
}
```
Thoughts?

scooter-dangle:
> ```rust
> pub struct TranslationTable {
>     table: HashMap<char, char>,
> } impl TranslationTable {
>     pub fn new(f: &str, t: &str) -> Result<Self, &'static str> {
>         if f.len() != t.len() {
>             Err("Arguments passed to maketrans must have equal length")
>         } else {
>             Ok(TranslationTable { table: HashMap::from_iter(f.chars().zip(t.chars())) })
>         }
>     }
> 
>     pub fn translate(&self, s: String) -> String {
>         s
>             .chars()
>             .map(|c| *self.table.get(&c).unwrap_or(&c))
>             .collect::<String>()
>     }
> }
> ```
> 
> Thoughts?

Very close. You'll want to check character length instead of byte length, since we're operating on characters. (You can use the `char_len` function in `main.rs` for that.)

Also, you'll want to tweak the error message to be 
```rust
"Arguments passed to `TranslationTable::new` must have equal character length"
```

charliethomson:
Donezo, anything else?

.gitignore
```diff
@@ -8,3 +8,9 @@ Cargo.lock
 
 # These are backup files generated by rustfmt
 **/*.rs.bk
+
+qepy.deroff
+pyparseqelectrotech.py
+main.py
+deroff.py
+.qelectrotech.1
```
pickfire: I believe these changes are not required.

charliethomson: You're right that was on my local one, oops. db0a659

src/util.rs
```diff
@@ -7,3 +7,82 @@ macro_rules! regex {
...
 }
+
+use crate::char_len;
+use std::{collections::HashMap, iter::FromIterator};
+
+pub struct TranslationTable {
+    table: HashMap<char, char>,
+}
+impl TranslationTable {
+    pub fn new(f: &str, t: &str) -> Result<Self, &'static str> {
```
pickfire: Could we make this `const fn` and use `&'static str`? I think we already know this at compile time. If we used these, we could just `debug_assert_eq!(char_len(f), char_len(t))`.

src/util.rs
```diff
@@ -7,3 +7,82 @@ macro_rules! regex {
...
+            Ok(TranslationTable {
+                table: HashMap::from_iter(f.chars().zip(t.chars())),
+            })
+        }
+    }
+
+    pub fn translate(&self, s: String) -> String {
+        s.chars()
+            .map(|c| *self.table.get(&c).unwrap_or(&c))
+            .collect::<String>()
```
pickfire:
```suggestion
            .collect()
```
I believed the type is already inferred?

src/util.rs
charliethomson: I don't think we could use `&'static str`'s because when you create the table (see below), you use refs of `self.tr_from` and `self.tr_to`.

(I have a (nearly) fully implemented version of deroff.rs that has `macro_tr` implemented, and I credited it to Anders, but I can't find where I got the code from :(, but this is from that (slightly modified to work with `TranslationTable`))
```rust
self.tr = match TranslationTable::new(&self.tr_from, &self.tr_to) {
    Ok(table) => Some(table),
    Err(e) => ..
};
```

src/util.rs
charliethomson: Yes, I don't know why I left that in

src/util.rs
pickfire: Yes, but looking at the code, `tr_from` and `tr_to` have essentially the same length, so in this case maybe we can just `assert!()` and let it panic since it does not have the same length?

charliethomson: I mean, we _could_, but I don't quite know why that'd be preferred over a `Result`
Am I missing something?
I feel like `Result` is clearer than a comment explaining that it _can_ panic

scooter-dangle: I think preserving a `Result` is fine for now. We can always `unwrap` directly if we want, and an invalid translation directive seems like something we should be able to handle without panicking.

src/util.rs
```diff
@@ -7,3 +7,82 @@ macro_rules! regex {
...
+        if char_len(f) != char_len(t) {
+            Err("Arguments passed to `TranslationTable::new` must have equal character length")
+        } else {
+            Ok(TranslationTable {
+                table: HashMap::from_iter(f.chars().zip(t.chars())),
+            })
+        }
+    }
+
+    pub fn translate(&self, s: String) -> String {
```
scooter-dangle: Should be able to take in a `&str` as input to avoid having to clone where we use it, no?
```suggestion
    pub fn translate(&self, s: &str) -> String {
```

src/deroff.rs
```diff
@@ -556,6 +557,23 @@ impl Deroffer {
...
+    /// if `self.tr` is set, instead of putting `s` into `self.output` directly,
+    /// it `translate`s it using the set translation table and puts the result
+    /// into `self.output`
+    fn condputs(&mut self, s: &str) {
+        let is_special =
+            { self.pic || self.eqn || self.refer || self.r#macro || self.inlist || self.inheader };
+
+        if !is_special {
+            if let Some(table) = &self.tr {
+                self.output.push_str(&table.translate(s.into()));
```
scooter-dangle: Can get rid of the conversion into a String if we allow `translate` to take a `&str`.
```suggestion
                self.output.push_str(&table.translate(s));
```

src/main.rs
```diff
@@ -1694,10 +1694,8 @@ mod tests {
     impl FileKind {
         fn content(self) -> &'static [u8] {
             match self {
-                FileKind::Good => {
-                    b"    \n#Hello, world!\
-                    \n#Hello, world! Autogenerated      "
-                }
+                FileKind::Good => b"    \n#Hello, world!\
+                    \n#Hello, world! Autogenerated      ",
```
scooter-dangle: I think maybe this is the result of a pre-1.40 formatter? Maybe an auto-formatter run by your editor?

charliethomson: Yeah, I was on 1.38.0. 6 week release cycle got me behind

PR #67, opened 2019-11-15
Implement the rest of the unclaimed `macro_<something>` functions

charliethomson:
This is all the macro functions not claimed or already implemented.
Please look at `macro_ds` (deroff.rs:741), it's super horrible but I couldn't find a less horrible way to do it.
edit: I am stupid and didn't look hard enough, so ^ this ^ is fixed

This needs `condputs` which needs `translate`, so it's built off my branch containing those implementations, but I can port it to another if we get a definite interface for them, thanks!
:smile:


src/deroff.rs
```diff
@@ -838,29 +838,18 @@ impl Deroffer {
...
-        let s = s.as_str();
-        self.s = self.skip_char(s, Some(2)).to_owned();
-        self.s = self.skip_leading_whitespace(s).to_owned();
-        while !self.s.is_empty() && Self::str_at(s, 0) != "\n" {
-            let c = Self::str_at(s, 0);
-            let mut ns = Self::str_at(s, 1);
-            self.s = self.skip_char(self.s.as_str(), Some(2)).to_owned();
+        let s = &self.s.clone();
+        // self.skip_char(2);
+        self.s = self.skip_char(&self.s, 2).to_owned();
```
pickfire: I believe this is outdated.

scooter-dangle: Hmm. We got rid of that clone before merging. Not sure why Github isn't labeling this code snippet as outdated.

PR #68, opened 2019-11-29
Implement font, numreq, var, size, esc, word, text, and esc_char_backslash

charliethomson:
I also had to add stubs for `comment`, `prch`, `text_arg`, `digit`, `condputs`, and `spec`.
I gave `Deroffer::reg_table` a type (`HashMap<String, String>`) based on the context it was used in.
I think that's about it.

edit: v this is talking about #69 v
Oh, I left commented out cleaner versions of a few lines that could be fixed with <insert future PR here> to change the interface of `str_at` and `skip_char` and probably also `is_white`, so yeah. Sorry for the PR spam Scott :P


src/deroff.rs
```diff
@@ -538,6 +538,307 @@ impl Deroffer {
...
+        unimplemented!()
+    }
+
+    fn var(&mut self) -> bool {
+
+        let mut reg = String::new();
+        let s0s1 = &self.s[0..2];
+        match s0s1 {
+            "\\n" => {
+                if let Some("dy") = self.s.get(3..5) {
```
scooter-dangle: No need for `let`
```suggestion
                if Some("dy") == self.s.get(3..5) {
```

src/deroff.rs
```diff
@@ -538,6 +538,307 @@ impl Deroffer {
...
+                    // self.skip_char(5);
+                    self.s = self.skip_char(self.s.as_str(), Some(5)).to_owned();
```
scooter-dangle: Nit: I think it would be more idiomatic here (and in subsequent lines) to prepend a `&` rather than append the `.as_str()`.
```suggestion
                    self.s = self.skip_char(&self.s, Some(5)).to_owned();
```

src/deroff.rs
```diff
@@ -538,6 +538,307 @@ impl Deroffer {
...
+        self.r#macro = m != 0;
+
+        true
+    }
+
+    fn text_arg(&self) -> bool {
+        unimplemented!()
+    }
+
+    fn var(&mut self) -> bool {
```
scooter-dangle: Yikes! This is a gross bear of a function! Nice work!

Need a test function with a decent number of assertions to show what it's doing and prevent us from breaking it in the future.

src/deroff.rs
charliethomson: Thank you for noticing, done:
e865801 and d6a76a7

src/deroff.rs
charliethomson: You're right, I should have looked at the `Deref` implementation, my mistake, fixed:
d2607a5

src/deroff.rs
charliethomson: This is the worst function in this entire file. I'm glad you said to go back and add tests. There's a solid amount of comments in the tests explaining what the code is doing, as well as a python version of the tests to run with the python source to make sure it perfectly mirrors the behavior. It's not perfect yet, `text_arg` needs to be implemented, but after that, we never have to look here ever again.
c40304f

src/deroff.rs
```diff
@@ -657,82 +657,91 @@ impl Deroffer {
         unimplemented!()
     }
 
+    fn prch(&self, idx: usize) -> bool {
+        self.s.get(idx..(idx+1))
```
pickfire: I believe the parenthesis is not needed here? Maybe we need to clippy clean the file later?

charliethomson: You're right, I should have done `(idx..=idx)`, I ran `fmt` and `clippy`: ea0c2fe

src/deroff.rs
```diff
@@ -538,12 +537,311 @@ impl Deroffer {
...
+            }
+            "[" => {
+                self.refer = true;
+                // self.condputs("\n");
+                return true;
+            }
+            "]" => {
+                self.refer = false;
+                // self.skip_char();
+                self.s = self.skip_char(&self.s, None).to_owned();
```
pickfire: Why do we have `to_owned()` for `String` everywhere? Isn't this way slower than parsing just `&str`?

charliethomson: `self.s` is a `String`. #69 exists, I don't like the interfaces for those functions because they make you convert into a `&str` and then back to `String`. Modifying self.s in place rather than converting it to a `&str` and then setting it to the result converted to a `String` would be significantly faster and cleaner

scooter-dangle: I was looking at this a little bit. I think there are tons of opportunities to get rid of some of the worst of the cloning. One thing to note for some of the cases that show up in this PR though, is that the signature of `condputs` from the perspective of this PR should likely be `condputs(&mut self, _: &str)` instead of `condputs(&self, _: &str)`, which will then prevent you from passing in an `&str` that is borrowed from `self`.

@pickfire, we might initially have something very slow by Rust standards. I'm _doubtful_ it will be any slower than the original—I'd expect it to be at least a little faster. However, we still don't have something fully working in the `deroff` module. I'd like to have that plus a comprehensive set of unit tests for `deroff`'s behavior before we revisit to optimize.

charliethomson: I should have known that, I feel like this will lead to more clones than before. I don't know how to handle stuff like in `word` without a clone

```rust
while let Some(m) = self.g_re_word.find(&self.s) {
    self.condputs(m.as_str());
```

pickfire: Yeah, if any I believe we can still discuss on the design of parsing later on. A rough idea is to just keep the original `String` or `&str` and use an index to iterate through it.

scooter-dangle:
We can also keep a reference to the original `str`, as part of the struct
definition (e.g., `struct Deroffer<'a> { output: Vec<Cow<'a, str>> /* Cow
so that we can own if we need to */, s: &'a str, ... }`). Not something we
need to do now, though.

charliethomson:
Is there anything I need to change on this PR?

pickfire:
I think we can just take in this first? By the way, I feel like refactoring how string parsing current works.

charliethomson:
> I think we can just take in this first? By the way, I feel like refactoring how string parsing current works.

I definitely agree, most of the code is kinda begging for a reformat.

pickfire:
Why is the original python code for this not removed? Is this still work in progress?

charliethomson:
@pickfire I literally never look at the source in the file I'm editing, so I didn't even know it was there, they're removed now

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
         unimplemented!()
     }
 
+    fn quoted_arg(&self) -> bool {
+        unimplemented!()
+    }
+
+    fn request_or_macro(&mut self) -> bool {
```
pickfire, line 545: Can the python code for this be deleted?

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+        // self.skip_char();
+        self.s = self.skip_char(&self.s, None).to_owned();
+
+        let s0 = &self.s[1..2];
```
pickfire, line 549: Possible out of bound error here, I believe we should use `self.s.bytes().get(1)` or something safer here.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+
+        match s0 {
+            "\\" => {
+                // if self.str_at(1) == Some('"') {
+                if Self::str_at(&self.s, 1) == "\"" {
+                    // self.condputs("\n");
```
pickfire, line 555: Why is this line being left out? It was in the original python code.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+
+    fn request_or_macro(&mut self) -> bool {
+        // self.skip_char();
+        self.s = self.skip_char(&self.s, None).to_owned();
+
+        let s0 = &self.s[1..2];
+
+        match s0 {
+            "\\" => {
+                // if self.str_at(1) == Some('"') {
```
pickfire, line 553: Is this still needed?

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                if Self::str_at(&self.s, 1) == "\"" {
+                    // self.condputs("\n");
+                    return true;
+                }
+            }
+            "[" => {
+                self.refer = true;
+                // self.condputs("\n");
```
pickfire, line 561: Why is this line being left out? It was in the original python code.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                return true;
+            }
+            "]" => {
+                self.refer = false;
+                // self.skip_char();
```
pickfire, line 566: Why is this line being left out? It was in the original python code.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                self.s = self.skip_char(&self.s, None).to_owned();
+                // return self.text();
```
pickfire, line 568: Why is this line being left out? It was in the original python code.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+            }
+            "." => {
+                self.r#macro = false;
+                // self.condputs("\n");
```
pickfire, line 572: Why is this line being left out? It was in the original python code.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                return true;
+            }
+            _ => (),
+        };
+
+        self.nobody = false;
+        let s0s1 = self.s[..2].to_owned();
```
pickfire, line 579: Possible out of bound panic here, I believe we should use `self.s.get(..2).unwrap_or_default()` or something similar here.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+        while !self.s.is_empty() && !Self::is_white(&self.s, 0) {
+            self.s = self.skip_char(&self.s, None).to_owned();
+        }
+
+        self.s = self.skip_leading_whitespace(&self.s).to_owned();
+
+        loop {
+            if !self.quoted_arg() && !self.text_arg() {
+                if !self.s.is_empty() {
+                    // self.condputs(self.str_at(0));
```
pickfire, line 599: I believe this is extra.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+            self.s = self.skip_char(&self.s, Some(m.end())).to_owned();
+            true
+        } else {
+            false
+        }
+    }
+
+    fn comment(&self) -> bool {
+        unimplemented!()
+    }
```
pickfire, line 620: I believe comment is already implemented right? Maybe you might need to rebase this with master?
https://github.com/rust-dc/fish-manpage-completions/blob/2e8c847b2bf0f04556126668aeef837425c34506/src/deroff.rs#L369-L374

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+
+    fn numreq(&mut self) -> bool {
+        // In the python, it has a check that is already handled in esc_char_backslash, which is
+        // the only place it gets called, so I'll omit that check here
+
+        // This is written as `self.macro += 1` in the source, but I dont know why
+        // it does the same thing (false -> true, true -> still true) :shrug:
+        // self.r#macro = true;
+        // Upon further investigation, this is the weirdest function ever
+        // This is just a state placeholder thing
```
pickfire, line 630: I believe you are missing a check here.
```python
        if self.str_at(1) in "hvwud" and self.str_at(2) == "'":
```
The `self.str_at(2)` part, yes the first part is duplicated but I don't see the second part being duplicated.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+        let mut m = self.r#macro as u8 + 1;
+
+        // self.skip_char(3);
+        self.s = self.skip_char(&self.s, Some(3)).to_owned();
+
+        // There's this, which has a comment explaining my thoughts right now
+        /*
+        while self.str_at(0) != "'" and self.esc_char():
+            pass  # Weird
```
pickfire, line 639: I believe this may still be needed to skip everything in between `'`, most like it put `Weird` because the syntax looks weird.

src/deroff.rs
pickfire: I believe this one can do a quick exit, like:
```rust
if self.str_at(2) != '\'' {
    return false;
}
```

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+        // And I dont know what the purpose of it would even be, if you can tell, lmk
+
+        // if self.str_at(0) == '\'' {
+        if Self::str_at(&self.s, 0) == "'" {
+            // self.skip_char(1);
+            self.s = self.skip_char(&self.s, None).to_owned();
+        }
+
+        m -= 1;
+        self.r#macro = m != 0;
```
pickfire, line 650: Can I know why is this needed? Shouldn't it be integer like the original source code?

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+
+        true
+    }
+
+    fn text_arg(&self) -> bool {
+        unimplemented!()
+    }
```
pickfire, line 657: This is also available in master.
https://github.com/rust-dc/fish-manpage-completions/blob/2e8c847b2bf0f04556126668aeef837425c34506/src/deroff.rs#L414-L436

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+
+    fn prch(&self, idx: usize) -> bool {
+        self.s
+            .get(idx..=idx)
+            .map(|c| !" \t\n".contains(c))
+            .unwrap_or_default()
```
pickfire, line 663: Should we parse just a single byte here instead of a string?

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+
+    // This function is the worst, there are a few comments explaining some of it in the test (test_var)
+    // its so hard to briefly put into words what this function does, basically depending on the state
+    // of self.s, it will either, change self.s to "", a part of self.s, or a value in self.reg_table
+    // which corresponds to a key that is part of self.s.
+    // This should be like 2 or 3 functions, but it's only one. So there's that. :-)
+    // NOTE: there is a call to text_arg that is commented out because it's not implemented, so the
+    // tests will need revised when it gets implemented
+    fn var(&mut self) -> bool {
+        let s0s1 = &self.s[0..2];
```
pickfire, line 674: This may fail, it should use something safer.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                    // self.skip_char(1);
+                    self.s = self.skip_char(&self.s, None).to_owned();
+                }
+                if let Some("]") = self.s.get(0..1) {
+                    // self.skip_char(1);
+                    self.s = self.skip_char(&self.s, None).to_owned();
+                } else {
+                    return false;
+                }
+            } else {
```
pickfire, line 724: I believe you may be missing
```python
            elif self.prch(2):
                reg = self.str_at(2)
                self.skip_char(3)
```

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                // This unwrap is safe because of the if
+                self.s = self.reg_table.get(&reg).unwrap().to_owned();
+                // This is unimplemented, so tests fail:
+                // self.text_arg();
+                return true;
+            } else {
+                return false;
+            }
+        } else {
+            return false;
```
pickfire, line 738: I believe these return could be without the `return` and `;`.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+        }
+    }
+
+    fn digit(&self, _: usize) -> bool {
+        unimplemented!()
+    }
```
pickfire, line 744: This is already implemented.
https://github.com/rust-dc/fish-manpage-completions/blob/2e8c847b2bf0f04556126668aeef837425c34506/src/deroff.rs#L407-L412

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+
+    fn size(&mut self) -> bool {
+        /* # We require that the string starts with \s */
```
pickfire, line 747: I think this can just use rust comment method. Best with `//` inside the function.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+            }
+            true
+        } else {
+            false
+        }
+    }
+
+    fn condputs(&self, _: &str) {
+        unimplemented!()
+    }
```
pickfire, line 763: This also have been implemented.
https://github.com/rust-dc/fish-manpage-completions/blob/2e8c847b2bf0f04556126668aeef837425c34506/src/deroff.rs#L711-L722

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+
+    fn esc(&mut self) -> bool {
+        /* # We require that the string start with backslash */
+        if let Some(c) = self.s.get(1..2) {
```
pickfire, line 767: I believe it may be better to just iterator one byte here rather than a string.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                // self.skip_char(idx);
+                if !self.esc_char_backslash() {
+                    // self.condputs(self.str_at(0));
+                    self.condputs(Self::str_at(&self.s, 0));
+                    // self.skip_char(1);
+                    self.s = self.skip_char(&self.s, None).to_owned();
+                }
+            } else {
+                self.condputs(&self.s);
+                self.s = String::new();
```
pickfire, line 813: Why not?
```rust
self.condputs(&self.s.drain(..).collect());
```

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                break;
```
pickfire, line 814: Why not just `break true` or `return true` here then remove the one below?

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+            }
+        }
+        got_something
+    }
+
+    fn text(&mut self) -> bool {
+        loop {
+            if let Some(idx) = self.s.find("\\") {
+                self.condputs(&self.s[..idx]);
+                // self.skip_char(idx);
```
pickfire, line 804: I believe we can keep this now, it is already here.

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+                "f" => self.font(),
+                "s" => self.size(),
+                "h" | "v" | "w" | "u" | "d" => self.numreq(),
+                "n" | "*" => self.var(),
+                "(" => self.spec(),
+                _ => self.esc(),
+            }
+        } else {
+            false
+        }
```
pickfire, line 837: Should take just a byte here? Can we directly just match that byte? Like
```rust
match self.s.bytes().nth(1) {
    Some('"') => self.comment(),
    ...
}
```

src/deroff.rs
```diff
@@ -538,12 +538,311 @@ impl Deroffer {
...
+    }
+
     fn not_whitespace(s: &str, idx: usize) -> bool {
         // # Note that this return False for the empty string (idx >= len(self.s))
         // ch = self.s[idx:idx+1]
         // return ch not in ' \t\n'
         // TODO Investigate checking for ASCII whitespace after mvp
-        s.get(idx..(idx + 1))
+        s.get(idx..=idx)
```
pickfire, line 845: Why not just check a byte?

src/deroff.rs
charliethomson: `quoted_arg` isn't implemented as of right now, but I removed the source for `request_or_macro`

src/deroff.rs
charliethomson: taken care of:
```rust
let s0 = self.s.chars().nth(1).unwrap_or('_'); // _ will be ignored by the match
```

note: I changed the type of the match from `&str` to `char` because it's only one character

src/deroff.rs
charliethomson: `condputs` is not implemented on this branch, so the tests fail with them uncommented. There are specific tests to compensate for this and also to test once `condputs` gets merged

src/deroff.rs
charliethomson: #75 isn't merged yet, once it is i can move this to the new interface

src/deroff.rs
charliethomson: See above

src/deroff.rs
charliethomson: It's the new interface's version of the code on the line below

src/deroff.rs
charliethomson: See replies about `condputs`

src/deroff.rs
charliethomson: See above

src/deroff.rs
charliethomson: taken care of:
```rust
let s0s1 = self.s.chars().take(2).collect::<String>();
```

src/deroff.rs
pickfire: Can you put a TODO here? At least we know we need can come back to check later.

pickfire:
@charliethomson Can you please help to take a look at the rest of the comments?

src/deroff.rs
pickfire: #75 is merged, I believe this patch needs a rebase.

scooter-dangle:
@charliethomson, there are conflicts with master. Can you resolve?

src/deroff.rs
charliethomson: fixed

src/deroff.rs
charliethomson: Fixing in the next commit

src/deroff.rs
charliethomson: Fixed

src/deroff.rs
charliethomson: Fixed

src/deroff.rs
charliethomson: we could go for a `char` instead of `str` but `u8`s are inadvisable.
revised code: 
```rust
self.s
    .chars()
    .nth(idx)
    .map(|c| !c.is_whitespace())
    .unwrap_or_default()
```

Note on the u8s thing: `char` is a type that wraps multi byte unicode code points, so if we went off of bytes, we could miss some non ascii (mutli byte) whitespace.

src/deroff.rs
charliethomson: Fixed:
```rust
let s0s1 = match self.s.get(0..2) {
    Some(s) => s,
    None => return false,
};
```

src/deroff.rs
```diff
@@ -376,18 +376,12 @@ impl Deroffer {
...
-        string
-            .char_indices()
-            .nth(idx)
-            .map(|(idx, charr)| &string[idx..(idx + charr.len_utf8())]) // Okay to directly index based on idx/charr construction.
+    fn str_at(&self, idx: usize) -> &str {
+        let s = &self.s;
+        s.char_indices()
+            .skip(idx)
+            .next()
+            .map(|(i, c)| &s[i..(i + c.len_utf8())])
```
pickfire: Do we even need to handle all UTF? I thought utf8 ascii position are still in place? I believe we are only checking those markups which is at most one usize.

src/deroff.rs
```diff
@@ -1011,15 +952,35 @@ impl Deroffer {
...
         // ch = self.s[idx:idx+1]
         // return ch not in ' \t\n'
         // TODO Investigate checking for ASCII whitespace after mvp
-        s.get(idx..=idx)
-            .map(|string| " \t\n".contains(string))
+        self.s
+            .get(idx..=idx)
+            .map(|string| !" \t\n".contains(string))
             .unwrap_or_default()
     }
```
pickfire, line 965:
```suggestion

    }
```

pickfire:
@charliethomson Please help to put the pull request as a draft, once you are done can ping me or make this a normal pull request so I can review one off. Thanks a lot.

src/deroff.rs
charliethomson: These have been fixed :)

src/deroff.rs
charliethomson: I'm not sure I follow, can you elaborate

src/deroff.rs
charliethomson: Fixed, added a comment explaining it as well

src/deroff.rs
charliethomson: Yes, thank you for noticing. Added it

src/deroff.rs
charliethomson: Fixed

src/deroff.rs
charliethomson: Fixed

src/deroff.rs
charliethomson: Fixed

src/deroff.rs
charliethomson: I'm not sure I understand what you mean

src/deroff.rs
charliethomson: Fixed

src/deroff.rs
charliethomson: `.bytes()` gets us `u8`s for each byte in the `String`, The example you have there doesn't compile. We could do `.chars()` to get the effect you're going for but I'm not a super big fan of allocating an entire iterator, going over a few elements, and then matching on an `Option` rather than just `if let`ting and then matching on a substring. I'll happily implement it your way if you'd prefer but I think it's a little misguided

src/deroff.rs
charliethomson: This is fixed:
```rust
self.s
    .chars()
    .nth(idx)
    .map(|op| !op.is_whitespace())
    .unwrap_or_default()```

File: src/deroff.rs (hunk 35 above)
Comment (charliethomson) [reply src/deroff.rs:969]: I don't think I understand, can you elaborate on this please?

File: src/deroff.rs (hunk 34 above)
Comment (charliethomson) [reply src/deroff.rs]: This was a @scooter-dangle thing. I just reformatted it to work in #75

**Comment by charliethomson** [comment]
@pickfire I fixed most things. There are a few I'm unsure about, I've replied to everything afaik. I'm not sure I understand the thing you pinged me in, can I get some clarification?

File: src/deroff.rs (hunk 29 above)
Comment (charliethomson) [reply src/deroff.rs:931]: Oh also, this
```
error[E0502]: cannot borrow `self.s` as mutable because it is also borrowed as immutable
   --> src/deroff.rs:933:32
    |
933 |                 self.condputs(&self.s.drain(..).collect::<String>());
    |                 ---- --------  ^^^^^^^^^^^^^^^^ mutable borrow occurs here
    |                 |    |
    |                 |    immutable borrow later used by call
    |                 immutable borrow occurs here
```

File: src/deroff.rs (hunk 32 above)
Comment (scooter-dangle) [reply src/deroff.rs:957]: @pickfire's example would work if you adjust the match arm to be `Some(b'"')` instead of `Some('"')`.

In either case (substring vs byte), his observation of matching on `Some` rather than extracting to avoid nesting the `match` in an `if let` makes sense—no need to make `None` a special case via an `else` when `match` already handles cases like that.

File: src/deroff.rs (hunk 7 above)
Comment (scooter-dangle) [reply src/deroff.rs:707]: Wait...is this actually resolved? Did you add the code you mention somewhere else?

File: src/deroff.rs (hunk 14 above)
Comment (scooter-dangle) [reply src/deroff.rs:735]: Same as previous comment...is this actually fixed?

File: src/deroff.rs [hunk 36]
```diff
@@ -917,6 +1142,203 @@ fn test_is_white() {
...
+    d.s = "\\*[foo bar :)".to_owned();
+    assert!(d.var() == false);
+    assert!(d.s == "");
+
+    // B not in reg_table
+    d.s = "\\*[foo bar]abcd".to_owned();
+    assert!(d.var() == false);
+    assert!(d.s == "abcd");
+
+    // Here's a python version of these tests
```
scooter-dangle, line 1250: Can remove all this commented-out code. Maybe something to keep on a separate branch?

src/deroff.rs
```diff
@@ -1121,40 +1543,39 @@ fn test_number() {
...
-//                 break
-//             else:
-//                 self.condputs(self.s[:idx])
-//                 self.skip_char(idx)
-//                 if not self.esc_char_backslash():
-//                     self.condputs(self.str_at(0))
-//                     self.skip_char()
-//         return True
+
+//     def digit(self, idx):
```
scooter-dangle, line 1550: Wasn't digit already implemented?

src/deroff.rs
charliethomson: Removed

src/deroff.rs
charliethomson: Yes, these always get me. Removed

src/deroff.rs
charliethomson: Actually fixed now

src/deroff.rs
charliethomson: Fixed

src/deroff.rs
charliethomson: Oops, fixed

src/deroff.rs
scooter-dangle: Per discussion at the meetup, the only difference in the byte vs char approach is that the `_` branch (that points to `self.esc()`) won't be triggered if we use `str::get` and the byte isn't a valid single-byte UTF-8 sequence, but _will_ be triggered if we don't require it to start and end correctly on UTF-8 boundaries. @charliethomson is doing some quick tests of the original Python to see if it can tell us one way or the other the desired behavior.

charliethomson: Fixed

src/deroff.rs
charliethomson: Fixed

src/deroff.rs
scooter-dangle: Not sure what this was. Resolving.

Issue #63, opened 2019-11-09
Changes to the interfaces for Deroffer str_at and skip_char to not be so ugly


charliethomson:
I think we should make `str_at` a method instead of an associated function, as well as, for `str_at` and `skip_char`, stop taking a &str argument and just modify `self.s` within the function.
I also think we should stop taking an `Option<usize>` on `skip_char` and just take a `usize` (from what I can gather, the `Option` is only used because the python uses a default arg and it wants to mirror that, however I feel that its clearer to just take a number)

# Problem 
```rust
if let Some("dy") = self.s.get(3..5) {
    // This
    self.s = self.skip_char(self.s.as_str(), Some(5)).to_owned();
    return true;
} else if Self::str_at(self.s.as_str(), 2) == "(" && self.prch(3) && self.prch(4) {
    // This
    self.s = self.skip_char(self.s.as_str(), Some(5)).to_owned();
    return true;
} else if Self::str_at(self.s.as_str(), 2) == "]" && self.prch(3) {
    // This
    self.s = self.skip_char(self.s.as_str(), Some(3)).to_owned();
    // This
    let mut s0 = Self::str_at(self.s.as_str(), 0);
    while !s0.is_empty() && s0 != "]" {
        // This
        self.s = self.skip_char(self.s.as_str(), None).to_owned();
        // This
        s0 = Self::str_at(self.s.as_str(), 0);
    }
    return true;
}
```

# Current interface
```rust
Deroffer::str_at(string: &str, idx: usize) -> &str

Deroffer::skip_char(&self, s: &str, amount: Option<usize>) -> &str
```

# Proposed change
```rust
Deroffer::str_at(&self, idx: usize) -> Option<char>

Deroffer::skip_char(&mut self, amount: usize) -> ()
```

# after
```rust
if let Some("dy") = self.s.get(3..5) {
    // self.s = self.skip_char(self.s.as_str(), Some(5)).to_owned();
    self.skip_char(5);
    return true;
} else if Self::str_at(self.s.as_str(), 2) == "(" && self.prch(3) && self.prch(4) {
    // self.s = self.skip_char(self.s.as_str(), Some(5)).to_owned();
    self.skip_char(5);
    return true;
} else if Self::str_at(self.s.as_str(), 2) == "]" && self.prch(3) {
    // self.s = self.skip_char(self.s.as_str(), Some(3)).to_owned();
    self.skip_char(3);
[...]

PR #69, opened 2019-11-29
Clean the interface for a bunch of functions

charliethomson:
#63 
You can see the possible effect of these changes if you look at `var` in #68, where I've commented out the "new" version right above the dirty version


src/deroff.rs
```diff
@@ -360,16 +360,33 @@ impl Deroffer {
         })
     }
 
-    fn skip_char<'a>(&self, s: &'a str, amount: Option<usize>) -> &'a str {
+    fn skip_char(&mut self, amount: usize) {
+        self.s = self.s.get(amount..).unwrap_or("").to_owned();
```
pickfire: I think it should not overwrite the original bytes.

src/deroff.rs
```diff
@@ -360,16 +360,33 @@ impl Deroffer {
...
+    }
+
+    /* fn skip_char<'a>(&self, s: &'a str, amount: Option<usize>) -> &'a str {
         let amount = amount.unwrap_or(1);
         s.get(amount..).unwrap_or("")
+    } */
+
+    fn skip_leading_whitespace(&mut self) {
+        self.s = self.s.trim_start().to_owned();
```
pickfire: Same for this.

src/deroff.rs
pickfire: Looking at python source code, this looks like what they are doing. I think it should be good then.

src/deroff.rs
```diff
@@ -383,15 +400,40 @@ impl Deroffer {
...
+    /* fn is_white<'a>(s: &'a str, idx: usize) -> bool {
+        s
+          .chars()  // Chars
+          .nth(idx) // Option<char>
+          .unwrap_or('a') // char
+          .is_whitespace() // bool
+    } */
+
+    // This is also known as `prch` apparently
+    fn not_whitespace(&self, idx: usize) -> bool {
```
pickfire: Why not rename this back to `prch`, the same as the original version?

src/deroff.rs
charliethomson: Yeah I think (based on the python) it should just move the string forward `amount` chars,
here's the source for reference:
```python
def skip_char(self, amt=1):
    self.s = self.s[amt:]
```

src/deroff.rs
charliethomson: Same for this one, in the python it just makes a call to `lstrip` which does the same as `trim_start`

src/deroff.rs
charliethomson: I implemented it separately in c40304f just to make sure my implementation didn't break anything, after some testing, my implementation was functionally equivalent to `not_whitespace`. I don't know if there's a reason it was renamed, (cc @scooter-dangle) but the only reason I didn't merge the functions in #68 is because its an associated function rather than a method, and this was the PR for fixing all of those

src/deroff.rs
```diff
@@ -383,15 +400,40 @@ impl Deroffer {
...
+        !" \t\n".contains(self.s.get(idx..idx+1).unwrap_or_default())
```
pickfire:
```suggestion
        self.s.get(idx..idx+1).map_or(false, |s| !" \t\n".contains(s))
```

src/deroff.rs
pickfire: Why not?
```suggestion
        self.s.drain(..amount);
```

charliethomson: I didn't know `drain` existed on `String` lol. I thought it was only on `Iterator`s but that's definitely clearer.

charliethomson:
@pickfire I actually don't know, I'll remove them now

src/deroff.rs
scooter-dangle: I think @PureW had brought it up (?) when we were looking at `prch` that the function name made no sense—especially when compared to some of the other functions like `is_white` and `digit`. We generally don't rename things as part of the initial port, but this one seemed especially egregious, so we made an exception. Helpful that you added the `prch` comment, @charliethomson, in case someone is looking for what happened to that function.

src/deroff.rs
```diff
@@ -365,51 +366,49 @@ impl Deroffer {
...
-            .unwrap_or_default()
+            .map(|(i, c)| &s[i..(i + c.len_utf8())])
+            .unwrap_or("")
     }
 
-    fn is_white<'a>(s: &'a str, idx: usize) -> bool {
-        match Self::str_at(s, idx) {
-            "" => false,
-            c => c.chars().all(|c| c.is_whitespace()),
-        }
```
pickfire: Why not reuse `str_at` the same way python does it?

https://github.com/rust-dc/fish-manpage-completions/pull/73/files#diff-5b0804df3203e0b4fc4f245618d61d53R399-R404
```rust
    fn is_white<'a>(&self, idx: usize) -> bool {
        match self.str_at(idx) {
            "" => false,
        }
    }
```

src/deroff.rs
```diff
@@ -455,115 +454,694 @@ impl Deroffer {
...
+            self.skip_char(1);
+        }
+
+        m -= 1;
+        self.r#macro = m != 0;
+
+        true
+    }
+
+    fn text_arg(&self) -> bool {
```
pickfire: Can you please help to add https://github.com/rust-dc/fish-manpage-completions/pull/73/files#diff-5b0804df3203e0b4fc4f245618d61d53L413-R437 as well if you are planning to rebase it with master, it was already fixed in master but I change it as well.

src/deroff.rs
```diff
@@ -455,115 +454,694 @@ impl Deroffer {
...
+            true
+        } else {
+            false
+        }
+    }
+
+    fn esc(&mut self) -> bool {
+        /* # We require that the string start with backslash */
+        // self.s.get(1..2).
+        match self.s.to_owned().get(1..2) {
```
pickfire: Why do we needa `to_owned()` here?

src/deroff.rs
```diff
@@ -455,115 +454,694 @@ impl Deroffer {
...
     }
 
-    fn not_whitespace(s: &str, idx: usize) -> bool {
-        // # Note that this return False for the empty string (idx >= len(self.s))
-        // ch = self.s[idx:idx+1]
-        // return ch not in ' \t\n'
-        // TODO Investigate checking for ASCII whitespace after mvp
-        s.get(idx..(idx + 1))
-            .map(|string| " \t\n".contains(string))
-            .unwrap_or_default()
```
pickfire: Maybe you could add a comment
```rust
    /// Like esc_char, but we know the string starts with a backslash
```
Based on https://github.com/fish-shell/fish-shell/blob/master/share/tools/deroff.py#L613

pickfire:
@charliethomson I sent you a pull request to remove some clone https://github.com/charliethomson/fish-manpage-completions/pull/1

src/deroff.rs
```diff
@@ -360,36 +360,34 @@ impl Deroffer {
...
-        string
-            .char_indices()
+    fn str_at(&mut self, idx: usize) -> &str {
+        let s = &self.s;
+        s.char_indices()
             .skip(idx)
             .next()
-            .map(|(idx, charr)| &string[idx..(idx + charr.len_utf8())]) // Okay to directly index based on idx/charr construction.
-            .unwrap_or_default()
+            .map(|(i, c)| &s[i..(i + c.len_utf8())])
```
pickfire: Can we keep `idx` to match the other parts?
```suggestion
            .map(|(idx, c)| &s[idx..(idx + c.len_utf8())])
```

charliethomson:
Somehow I completely broke this fork, I'm gonna work on fixing it 😞

charliethomson:
So, what happened is, I merged this into my master instead of pulling it in. So, I now have all my commits on this branch (it may have happened to another one, I'll investigate in a sec). I made a new branch ([deuglify-fix](https://github.com/charliethomson/fish-manpage-completions/tree/deuglify-fix)) with only the code that is supposed to be on this branch

charliethomson:
I decided it was better to just let this branch die. I moved all the actually relevant code to #75, and I'll close this PR

PR #71, opened 2019-12-05
ci: drop travis-ci in favor of github action for now

scooter-dangle:
Might re-explore in the future. For now our needs are pretty simple, and
Github Actions seem a smidge faster (although could be due to
misconfiguration on my part).

Note: subsumes #70


scooter-dangle:
womp womp...built-in environment is using an out-dated version of Rust, which breaks the formatting check.

PR #72, opened 2020-01-04
feat: add type1 fallback1, fallback2

pickfire:
`add_diagnostic` not added as does not work globally and `Type1` cannot access that method, not sure how to solve this with the current `add_diagnostic`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,6 +8,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use itertools::Itertools;
+use regex::Regex;
 use structopt::StructOpt;
 
 #[cfg(test)]
@@ -157,9 +158,9 @@ impl App {
 //         data = data[1:len(data)-1]
 //     return data
 
-fn unquote_double_quotes(data: String) -> String {
+fn unquote_double_quotes(data: &str) -> &str {
     if data.len() > 2 && (data.as_bytes()[0], *data.as_bytes().last().unwrap()) == (b'"', b'"') {
-        data[1..(data.len() - 1)].into()
+        &data[1..(data.len() - 1)]
     } else {
         data
     }
@@ -172,15 +173,15 @@ fn unquote_double_quotes(data: String) -> String {
 //         data = data[1:len(data)-1]
 //     return data
 
-fn unquote_single_quotes(data: String) -> String {
+fn unquote_single_quotes(data: &str) -> &str {
     if data.len() > 2 && (data.as_bytes()[0], *data.as_bytes().last().unwrap()) == (b'\'', b'\'') {
-        data[1..(data.len() - 1)].into()
+        &data[1..(data.len() - 1)]
     } else {
         data
     }
 }
 
-fn fish_escape_single_quote(string: String) -> String {
+fn fish_escape_single_quote(string: &str) -> String {
     format!("'{}'", string.replace(r"\", r"\\").replace(r"'", r"\'"))
 }
 
@@ -254,7 +255,7 @@ fn fish_options(options: &str, existing_options: &mut HashSet<String>) -> Vec<St
             fish_opt,
             // Direct indexing of `option` won't panic due to how `num_dashes`
             // is calculated. (I promise!)
-            fish_escape_single_quote(String::from(&option[num_dashes..]))
+            fish_escape_single_quote(&option[num_dashes..])
         );
 
         if existing_options.insert(option.clone()) {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -157,9 +158,9 @@ impl App {
 //         data = data[1:len(data)-1]
 //     return data
 
-fn unquote_double_quotes(data: String) -> String {
+fn unquote_double_quotes(data: &str) -> &str {
```
scooter-dangle: :+1:

src/main.rs
```diff
@@ -585,80 +586,100 @@ impl ManParser for Type1 {
...
-        unimplemented!()
+    fn fallback(
+        &self,
+        mut options_section: &str,
+        built_command_output: &mut Vec<String>,
+        existing_options: &mut HashSet<String>,
+        cmd_name: &str,
+    ) -> bool {
+        // add_diagnostic("Trying fallback");
+        let options_parts_re = Regex::new(r"\.TP( \d+)?((?s:.)*?)\.TP").unwrap();
```
scooter-dangle: Should default to using `regex!` macro in case function ends up being called more than once per program run (either now or in a future refactor).

```suggestion
        let options_parts_re = regex!(r"\.TP( \d+)?((?s:.)*?)\.TP");
```

src/main.rs
```diff
@@ -585,80 +586,100 @@ impl ManParser for Type1 {
...
+                built_command(
+                    option_name,
+                    option_desc.as_str(),
+                    built_command_output,
+                    existing_options,
+                    cmd_name,
+                );
+            } else {
+                // add_diagnostic(format!("{:?} does not contain '-'", option_name));
+                return false;
```
scooter-dangle: Hmm. Why doesn't the original enter an infinite loop in this case, since it doesn't return on this branch? Is there mutation in the Python source that I'm not recognizing?

src/main.rs
```diff
@@ -585,80 +586,100 @@ impl ManParser for Type1 {
...
+        &self,
+        options_section: &str,
+        built_command_output: &mut Vec<String>,
+        existing_options: &mut HashSet<String>,
+        cmd_name: &str,
+    ) -> bool {
+        // add_diagnostic("Trying last chance fallback");
+        let ix_remover_re = Regex::new(r"\.IX.*").unwrap();
+        let trailing_num_re = Regex::new(r"\d+$").unwrap();
+        let options_parts_re = Regex::new(r"\.IP ((?s:.)*?)\.TP").unwrap();
```
scooter-dangle: typo?
```suggestion
        let options_parts_re = Regex::new(r"\.IP ((?s:.)*?)\.IP").unwrap();
```

src/main.rs
pickfire: Yeah, I did saw that macro but clippy complains about it so I am not sure to include it.

src/main.rs
pickfire: Yup, nice findings. I probably just copied the one above.

src/main.rs
pickfire: Oh, yes. I probably got confused with https://github.com/rust-dc/fish-manpage-completions/pull/72/files#diff-639fbc4ef05b315af92b4d836c31b023L644 and https://github.com/rust-dc/fish-manpage-completions/pull/72/files#diff-639fbc4ef05b315af92b4d836c31b023L653.

Yeah, it should not return. In fact, this block is not required without the `add_diagnostic`.

pickfire:
@scooter-dangle Thanks for reviewing. I have updated the issues you mentioned.

src/main.rs
```diff
@@ -585,80 +586,99 @@ impl ManParser for Type1 {
...
-//         return True
+    fn fallback2(
+        &self,
+        options_section: &str,
+        built_command_output: &mut Vec<String>,
+        existing_options: &mut HashSet<String>,
+        cmd_name: &str,
+    ) -> bool {
+        // add_diagnostic("Trying last chance fallback");
+        let ix_remover_re = Regex::new(r"\.IX.*").unwrap();
```
scooter-dangle:
```suggestion
        let ix_remover_re = regex!(r"\.IX.*");
```

src/main.rs
```diff
@@ -585,80 +586,99 @@ impl ManParser for Type1 {
...
+        let trailing_num_re = Regex::new(r"\d+$").unwrap();
```
scooter-dangle:
```suggestion
        let trailing_num_re = regex!(r"\d+$");
```

src/main.rs
```diff
@@ -585,80 +586,99 @@ impl ManParser for Type1 {
...
+        let options_parts_re = Regex::new(r"\.IP ((?s:.)*?)\.IP").unwrap();
```
scooter-dangle:
```suggestion
        let options_parts_re = regex!(r"\.IP ((?s:.)*?)\.IP");
```

PR #74, opened 2020-01-06
fix: Weird encoding on fixtures/qelectrotech.1

charliethomson:
There was something fishy going on with qelectrotech.1. It wasn't encoding the unicode characters correctly from the looks, see examples below

Expected: QElectroTech est un éditeur de schémas électriques. Les schémas (*.qet) et les éléments électriques (*.elmt) sont enregistrés au format XML.
Got: QElectroTech est un �diteur de sch�mas �lectriques. Les sch�mas (*.qet) et les �l�ments �lectriques (*.elmt) sont enregistr�s au format XML.

Expected: .TH QELECTROTECH 1 "AOÛT 2008" QElectroTech "Manuel utilisateur"
Got: .TH QELECTROTECH 1 "AO�T 2008" QElectroTech "Manuel utilisateur"

It doesn't show up on github unless you look at the raw, but if you `clone` you get the weird encoding and it was breaking some of my tests, so I figured I'd fix it

Here are some links to the raw versions of [mine](https://raw.githubusercontent.com/charliethomson/fish-manpage-completions/qelectrotech/fixtures/qelectrotech.1) and the [current one](https://raw.githubusercontent.com/charliethomson/fish-manpage-completions/master/fixtures/qelectrotech.1)

edit: I put the same link twice :)

### DIFF
--- a/fixtures/qelectrotech.1
+++ b/fixtures/qelectrotech.1
@@ -1,6 +1,6 @@
-.TH QELECTROTECH 1 "AOÛT 2008" QElectroTech "Manuel utilisateur"
+.TH QELECTROTECH 1 "AOÃ›T 2008" QElectroTech "Manuel utilisateur"
 .SH NOM
-qelectrotech \- Éditeur de schémas électriques
+qelectrotech \- Ã‰diteur de schÃ©mas Ã©lectriques
 .SH SYNOPSIS
 .B qelectrotech
 .B [\-\-common\-elements\-dir\fR=\fIREP\fB]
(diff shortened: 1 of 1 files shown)


scooter-dangle:
It _is_ an annoying encoding—it's [latin1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1)! Unfortunately we'll need to be able to handle it to not break compatibility with the original program. :disappointed:

charliethomson:
It doesn't seem to be latin1. I copied the correct version (from github, the visualised one), and ran it through [this](https://www.mobilefish.com/services/latin_utf_base64_to_hex/latin_utf_base64_to_hex.php) latin1 to hex converter and got [this diff](https://www.diffchecker.com/vciwKH5O), the left is the output of the converter and the right is the raw bytes from the file (direct clone from rust-dc/master). In addition, it seems wrong to begin with because most of the latin1 characters (é, î, etc) are 3 bytes rather than 1. I may be wrong, but it could be github or git breaking something, I don't really know. It just doesn't seem right.

oh, P.S, I tested deroffing it with the python and it kept the borked characters (didn't """parse""" them correctly. But when I swapped in mine it worked fine. 🤷‍♂

scooter-dangle:
> It doesn't seem to be latin1. I copied the correct version (from github, the visualised one)

You'd probably not want to copy that version when determining encoding. Github appears to be automatically detecting and converting the encoding to UTF-8 for more convenient display.

I saw initially that it was Latin1 due to Vim reporting that as the encoding in my status bar when I opened the file. We can also make a really small and naïve program for doing this kind detection using the `chardet` crate:

```sh
cargo new --bin chardet
cd chardet
echo 'chardet = "0.2"' >> Cargo.toml
```

and then put the following in `src/main.rs`:

```rust
use std::io::{self, Read};

fn main() {
    let mut buffer = vec![];
    io::stdin().read_to_end(&mut buffer).unwrap();
    let (encoding, confidence, _) = chardet::detect(&buffer);
    println!("encoding: {} ({})", encoding, confidence);
}
```

The program will load everything from stdin into a byte-vector, run `chardet`'s detection function, and then print the detected encoding and level of confidence in that detection.

I have it in a sibling directory to `fish-manpage-completions`, so I can do the following:

```
~/codes/chardet (master|…)> cat ../fish-manpage-completions/fixtures/qelectrotech.1 | cargo run
   Compiling chardet v0.1.0 (/home/scott/codes/chardet)
    Finished dev [unoptimized + debuginfo] target(s) in 1.07s
     Running `target/debug/chardet`
encoding: ISO-8859-1 (0.73)
```

Which gives us ISO-8859-1, aka latin1.

scooter-dangle:
> In addition, it seems wrong to begin with because most of the latin1 characters (é, î, etc) are 3 bytes rather than 1.

So...those are multi-byte characters when encoded in UTF-8. In latin1, they're each encoded with only a single byte—bytes which don't map to anything if you try to interpret them as UTF-8, hence the [� replacement character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character).

> oh, P.S, I tested deroffing it with the python and it kept the borked characters (didn't """parse""" them correctly. But when I swapped in mine it worked fine. :man_shrugging:

That makes sense given the current state of the code. We have yet to port the code that attempts to decode the input as latin1.

- - -

The fun of string encodings!

PR #76, opened 2020-01-10
esc_char_backslash

sstanfield:
This one is pretty trivial and mostly uses functions Charlie is porting.  The only real content is esc_char_backslash.

### DIFF
--- a/src/deroff.rs
+++ b/src/deroff.rs
@@ -616,26 +616,152 @@ impl Deroffer {
         write.flush().unwrap()
     }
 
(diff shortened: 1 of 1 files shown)


pickfire:
I believe you are modifying way too many changes, I think we can keep the original python source code for reference until the rust code was implemented and delete it.

Can you please reduce the changes?

sstanfield:
@pickfire Sorry for the delay and you are correct about this being messy.  I kept the python source but I moved it around which made the PR a lot more complex then needed,  It has been overcome by events though (Charlie"s work) so I am just going to close this PR.

PR #77, opened 2020-01-10
feat: add Type1::Manparser

pickfire:


src/main.rs
```diff
@@ -535,79 +535,80 @@ impl ManParser for Type1 {
         manpage.contains(r#".SH "OPTIONS""#)
     }
 
-    fn parse_man_page(&mut self, manpage: &str) -> Option<String> {
+    fn parse_man_page(&mut self, manpage: &str, cmd_name: &str) -> Option<String> {
+        let options_section_re = regex!(r#"\.SH "OPTIONS"((?s:.)*?)(\.SH|\z)"#);
+        let options_section_matched = options_section_re.find(manpage);
+
+        let mut options_section = options_section_matched.unwrap().as_str();
```
charliethomson: Why not
```suggestion
        let mut options_section = &options_section_matched.unwrap_or_default()
```

src/main.rs
```diff
@@ -535,79 +535,80 @@ impl ManParser for Type1 {
...
+            let last_dotpp_index = data.rfind(".PP");
+            if let Some(idx) = last_dotpp_index {
+                data = &data[idx + 3..];
+            }
+
+            let data = remove_groff_formatting(data);
+            let data: Vec<_> = data.split(".RS 4").collect();
+            if data.len() > 1 {
+                let option_name = data[0].trim();
+                if option_name.find('-').is_some() {
```
charliethomson: Why not
```suggestion
                if option_name.contains('-') {
```

src/main.rs
```diff
@@ -535,79 +535,80 @@ impl ManParser for Type1 {
...
+                    let option_desc = data[1].trim().replace('\n', " ");
+                    built_command(
+                        option_name,
+                        option_desc.as_str(),
+                        &mut built_command_output,
+                        &mut existing_options,
+                        cmd_name,
+                    );
+                } else {
+                    // add_diagnostic(format!("{:?} doesn't contain '-'", option_name));
```
charliethomson: I think we need to decide how functions will have access to an `App` and its methods. Thanks for putting these in here now though

src/main.rs
pickfire:
```rust
error[E0599]: no method named `unwrap_or_default` found for type `std::option::Option<regex::re_unicode::Match<'_>>` in the current scope
   --> src/main.rs:542:60
    |
542 |         let mut options_section = &options_section_matched.unwrap_or_default();
    |                                                            ^^^^^^^^^^^^^^^^^ method not found in `std::option::Option<regex::re_unicode::Match<'_>>`
    |
    = note: the method `unwrap_or_default` exists but the following trait bounds were not satisfied:
            `regex::re_unicode::Match<'_> : std::default::Default`
```

src/main.rs
pickfire: Thanks, nice findings. I forgot about that.

src/main.rs
charliethomson: Ah, makes sense. I thought it was a string lol

src/main.rs
pickfire: I think using `log` would be better for this. I talked to @scooter-dangle about it, looks like two popular crate that we need to use are either `simple_logger` and `env_logger` (which I prefer).

pickfire:
@charliethomson Any news on this?

pickfire:
@charliethomson Any news on this?

pickfire:
Anything else that I could do here?

PR #79, opened 2020-01-12
Implement 'number' (baby steps)

rschmertz:
Implemented what will be the first of a number of pretty similar functions returning an Option<& str>. If it checks out, similar ones (word, others) should be straightforward.


src/deroff.rs
```diff
@@ -55,7 +55,7 @@ impl Deroffer {
     fn new() -> Deroffer {
         Deroffer {
             g_re_word: crate::regex!(r##"[a-zA-Z_]+"##),
-            g_re_number: crate::regex!(r##"[+-]?\d+"##),
+            g_re_number: crate::regex!(r##"^[+-]?\d+"##),
```
pickfire: Why do we need the `^` change? The original version uses the same as the current one. https://github.com/fish-shell/fish-shell/blob/master/share/tools/deroff.py#L268

src/deroff.rs
```diff
@@ -761,7 +761,14 @@ impl Deroffer {
     //             return self.esc()
 
     fn number<'a>(&mut self, s: &'a str) -> Option<&'a str> {
```
pickfire: Why not just return a `bool` like https://github.com/fish-shell/fish-shell/blob/master/share/tools/deroff.py#L603-L610? And I believe it should not take in `s` as argument.

It would be enough to just skip the numbers like the rest. The `unimplemented!()` could be removed in this case. It would also be better to use `mat` instead of `possible` like how the `regex` crate does (and most of how it was implemented in other parts).

src/deroff.rs
scooter-dangle: I think that it has to have the anchor based on the definition of Python's [`re.Pattern.match`](https://docs.python.org/3/library/re.html#re.Pattern.match), which unintuitively requires a match at the start of the string.

rschmertz: Yes, that's it. That, and the correct functioning of the parser seems to depend on this behavior of https://docs.python.org/3/library/re.html#re.Pattern.match. Let me know if you think otherwise.

pickfire: Oh, I did not know `match` is the *beginning of string*. Then `g_re_word`, `g_re_not_backslash_or_whitespace` and `g_re_font` also needs some changes.

src/deroff.rs
```diff
@@ -761,7 +761,14 @@ impl Deroffer {
...
-        unimplemented!()
+        //unimplemented!()
```
pickfire: I believe this line is not needed anymore.

src/deroff.rs
pickfire: Any news on this?

rschmertz: Sorry I have taken so long to get back to this.

Regarding the function signature, I'd have to defer to others; I didn't make the choice. There seems to be a set of functions of this style, and another set of functions which have signatures which return bool, in the same way the original Python methods do.

For the other parts of this comment, could I ask you to break up into more parts to help make it clearer what you're talking about?

src/deroff.rs
rschmertz: I've removed it, thanks.

src/deroff.rs
pickfire: Try to use `bool` since we are starting to make patches to change the `Option<&'a str>` to `bool`.

src/deroff.rs
```diff
@@ -761,7 +761,13 @@ impl Deroffer {
     //             return self.esc()
 
     fn number<'a>(&mut self, s: &'a str) -> Option<&'a str> {
-        unimplemented!()
+        let possible = self.g_re_number.find(s);
+        if let Some(m) = possible {
```
pickfire: Can you please rename `m` to `mat` to be congruent to regex crate?

src/deroff.rs
pickfire: I guess we can just merge this first and fix it later.

pickfire:
@rschmertz I have updated the patch to use `bool` and resolve the merge conflict. Thanks for sending in the patch.

PR #80, opened 2020-02-14
Remove str_eq

benjaminrsherman:


src/deroff.rs
```diff
@@ -840,6 +840,22 @@ impl Deroffer {
             None => panic!("do_line` called with empty string as argument"),
         }
     }
+
+    fn str_eq(&self, offset: usize, other: &str, len: usize) -> bool {
```
scooter-dangle: Can remove this guy now. :disappointed: womp womp

benjaminrsherman: Removed.

pickfire: This guy is not used in other functions.

pickfire:
@benjaminrsherman Thanks for sending in the patch but the function is not used anymore, so let's close this.

PR #81, opened 2020-02-20
Types

pickfire:


pickfire:
Yes, using `log` would be an option, also we need to remove those indentation things when logging this time too. We also need to take a look at some progress bar, [`indicatif`](https://crates.io/crates/indicatif) would be good for now.

pickfire:
@charliethomson Is @scooter-dangle still alive? Should we wait?

pickfire:
@charliethomson Nice, @scooter-dangle gave me write access but I still need an approval from someone with write access too. T_T

src/main.rs
```diff
@@ -484,28 +496,11 @@ fn test_complete_command() {
...
-        //"\&",
-        //"\f",
-    ] {
-        data = data.replace(marker, "");
-    }
-    // See note above about anomalous backslash
-    //data = data.replace("\-", "-");
-    let data = regex!(r##".PD( \d+)"##).replace_all(&data, "");
-    data.to_string()
+fn remove_groff_formatting(data: &str) -> Cow<'_, str> {
```
scooter-dangle, line 499:
```suggestion
fn remove_groff_formatting(data: &str) -> Cow<str> {
```

src/main.rs
```diff
@@ -484,28 +496,11 @@ fn test_complete_command() {
...
+    // using regex is twice as fast as manual replace
+    let re =
```
scooter-dangle, line 501: Are we sure this is strictly equivalent? I _hope_ the original wasn't relying on the explicit ordering, but we would want to verify that. Here's a contrived example of a string that the original would treat differently than a regex replace-all: `r###"\\fIfP"###`. The original will transform that to `""`, but the regex replace all will only do a single pass, which will result in `r###"\fP"###`.

Maybe retain some of the comments from original and a TODO to revisit?

src/main.rs
```diff
@@ -484,28 +496,11 @@ fn test_complete_command() {
...
+        regex!(r"\\fI|\\fP|\\f1|\\fB|\\fR|\\e|\.BI|\.BR|0\.5i|\.rb|\\^|\{ | \}|\.B|\.I|.PD( \d+)");
```
scooter-dangle, line 502: The `^` in the regex should have its own `\` to avoid being interpreted as a start of string anchor.

src/main.rs
pickfire: Thanks, I forgot about that type elision.

src/main.rs
pickfire: Ah, nice findings, I think I miss that.

src/main.rs
pickfire: Okay, but what I did is to replace all, the original regex wasn't doing that.

pickfire:
@scooter-dangle I have added the previous comments.

charliethomson:
@scooter-dangle bump

pickfire:
@charliethomson This was added with your `TypeDarwin` patch, can you also help me to review `Type3::ManParger`?

CC @scooter-dangle

src/main.rs
```diff
@@ -535,92 +550,77 @@ impl ManParser for Type1 {
...
+            }
+
+            let data = remove_groff_formatting(data);
+            if let Some(data) = data.splitn(2, ".RS 4").next_tuple::<(_, _)>() {
+                let option_name = data.0.trim();
+                if option_name.contains('-') {
+                    let option_name = unquote_double_quotes(option_name);
+                    let option_name = unquote_single_quotes(option_name);
+                    let option_desc = data.1.trim().replace('\n', " ");
+                    completions.add(option_name, &option_desc);
```
scooter-dangle, line 584: Could eagerly extract tuple contents so reader has one fewer intermediate binding to remember.
```suggestion
            if let Some((option_name, option_desc)) = data.splitn(2, ".RS 4").next_tuple::<(_, _)>() {
                let option_name = option_name.trim();
                if option_name.contains('-') {
                    let option_name = unquote_double_quotes(option_name);
                    let option_name = unquote_single_quotes(option_name);
                    let option_desc = option_desc.trim().replace('\n', " ");
                    completions.add(option_name, &option_desc);
```

src/main.rs
```diff
@@ -690,49 +680,46 @@ impl ManParser for Type2 {
...
+            let data = mat.get(3).unwrap().as_str();
+            let data = remove_groff_formatting(data);
+            let data = data.splitn(2, '\n').next_tuple::<(_, _)>();
+            if let Some(data) = data.filter(|data| !data.1.trim().is_empty()) {
+                let option_name = data.0.trim();
+                if option_name.contains('-') {
+                    let option_name = unquote_double_quotes(option_name);
+                    let option_name = unquote_single_quotes(option_name);
+                    let option_desc = data.1.trim().replace('\n', " ");
+                    completions.add(option_name, &option_desc);
```
scooter-dangle, line 708:
```suggestion
            if let Some((option_name, option_desc)) = data.filter(|(_, desc)| !desc.trim().is_empty()) {
                let option_name = option_name.trim();
                if option_name.contains('-') {
                    let option_name = unquote_double_quotes(option_name);
                    let option_name = unquote_single_quotes(option_name);
                    let option_desc = option_desc.trim().replace('\n', " ");
                    completions.add(option_name, &option_desc);
```

src/main.rs
```diff
@@ -535,92 +550,77 @@ impl ManParser for Type1 {
         manpage.contains(r#".SH "OPTIONS""#)
     }
 
-    fn parse_man_page(&mut self, manpage: &str) -> Option<String> {
-        unimplemented!();
+    fn parse_man_page(&self, manpage: &str, cmdname: &str) -> Option<String> {
+        let options_section_re = regex!(r#"\.SH "OPTIONS"((?s:.)*?)(\.SH|\Z)"#);
```
scooter-dangle, line 554: The `Type2` equivalent regex also uses `\Z`, but the `Type3` and `Type4` equivalents use `\\Z`. Is that intentional?

If possible, would be good to use the same style of `r#"..."#` or `r"..."` for all 4 so that the intentional differences between the regexes are more obvious to the reader.

src/main.rs
```diff
@@ -741,49 +728,49 @@ impl ManParser for Type3 {
...
+            if data.len() < 2 {
+                // add_diagnostic("Unable to split option from description");
+                return None;
+            }
+            let option_name = data[0].trim();
+            if option_name.contains('-') {
+                let option_name = unquote_double_quotes(option_name);
+                let option_name = unquote_single_quotes(option_name);
+                let option_desc = data[1].trim().replace("\n", " ");
+                completions.add(&option_name, &option_desc);
```
scooter-dangle, line 762: I _think_ this would work but haven't tested it locally. Wanted to mimic the `next_tuple` pattern you had used elsewhere.
```suggestion
            let (option_name, option_desc) = match data.splitn(2, '\n').next_tuple() {
                Some(tuple) => tuple,
                None => {
                    // add_diagnostic("Unable to split option from description");
                    return None;
                }
            };
            let option_name = option_name.trim();
            if option_name.contains('-') {
                let option_name = unquote_double_quotes(option_name);
                let option_name = unquote_single_quotes(option_name);
                let option_desc = option_desc.trim().replace("\n", " ");
                completions.add(&option_name, &option_desc);
```

src/main.rs
```diff
@@ -792,51 +779,49 @@ impl ManParser for Type4 {
...
+            let data = remove_groff_formatting(data);
+            let data = data.trim().splitn(2, '\n').next_tuple::<(_, _)>();
+
+            if let Some(data) = data {
+                let option_name = data.0.trim();
+                if option_name.contains('-') {
+                    let option_name = unquote_double_quotes(option_name);
+                    let option_name = unquote_single_quotes(option_name);
+                    let option_desc = data.1.trim().replace('\n', " ");
+                    completions.add(option_name, &option_desc);
```
scooter-dangle, line 808: Mimicking the pattern-match pattern mentioned elsewhere.

```suggestion
            if let Some((option_name, option_desc)) = data.trim().splitn(2, '\n').next_tuple() {
                let option_name = option_name.trim();
                if option_name.contains('-') {
                    let option_name = unquote_double_quotes(option_name);
                    let option_name = unquote_single_quotes(option_name);
                    let option_desc = option_desc.trim().replace('\n', " ");
                    completions.add(option_name, &option_desc);
```

src/main.rs
pickfire: Ops, `\Z` is the wrong one, I fixed it.
```suggestion
        let options_section_re = regex!(r#"\.SH "OPTIONS"((?s:.)*?)(\.SH|\\Z)"#);
```

pickfire: I do want to keep consistency between `r#"..."#` or `r"..."`. But the main reason `r#"..."#` was used is because there is `"` inside.

src/main.rs
pickfire: Oh, I did try that. The reason why I did not used that method is because it is longer, 2-3 lines last time I tried.

pickfire:
@scooter-dangle Thanks for reviewing the pull request. :D

Some of the suggestions that you have sent I did know them but I did not use them because they need to split into multiple lines after `cargo fmt` since they became too long. It also renders them less readable so I did not format them that way.

> Question: Are any of these functions testable individually?

Yes, but my plan is to not put tests here. I would go for fuzzing and take all my documentation, use the original parser and then compare it, I believe that approach would have a higher coverage.

What I am striving right now is to go for feature parity with the original python code.

src/main.rs
scooter-dangle: Ah, good point. I missed that. No need to change the quote style.

PR #86, opened 2020-04-12
reformat util:test_translationtable_new

charliethomson:
Use `hash_map!` from `common_macros` to remove some HashMap initialization clutter

+ Old: 
```rust
    let mut expected = HashMap::new();
    expected.insert('a', 'c');
    expected.insert('b', 'd');
    expected.insert('!', '.');
```

+ New:
```rust
    let expected = hash_map! {
        'a' => 'c',
        'b' => 'd',
        '!' => '.',
    };
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,4 @@ libflate = "*"
 
 [dev-dependencies]
 pretty_assertions = "*"
+common_macros = "*"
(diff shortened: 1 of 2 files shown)


pickfire:
I don't see how is this useful, it's the same amount of lines. @charliethomson Did the number of dependency crates or compilation time increase?

PR #87, opened 2020-04-26
Impl TypeDarwin parse_man_page

charliethomson:


pickfire:
@charliethomson Is it possible to build after my PR in `types` and send me a pull request in https://github.com/pickfire/fish-manpage-completions to `types` branch, that way I can include that in another #81 pull request too.

pickfire:
@charliethomson I added this to #81

PR #92, opened 2020-05-30
refactor: extract description with iterator

pickfire:
cc @charliethomson

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -855,23 +855,15 @@ impl ManParser for TypeDarwin {
             let name = line.split_whitespace().next().unwrap();
 
             // Extract the description
-            let mut desc_lines = Vec::new();
-            while let Some(line) = lines.next() {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -855,23 +855,15 @@ impl ManParser for TypeDarwin {
...
-                        desc_lines.push(line);
-                    }
-                }
-            }
-            let desc = desc_lines.join(" ");
+            let desc = lines
+                .by_ref()
+                .take_while(|line| Self::is_option(line))
+                .filter(|line| line.starts_with(".") && !line.starts_with(".\"")) // Ignore comments
+                .map(|line| Self::groff_replace_escapes(line))
```
scooter-dangle, line 862: Pointless nit: 
```suggestion
                .map(Self::groff_replace_escapes)
```

PR #93, opened 2020-06-11
feat: Add CellMap

charliethomson:
This adds a trait `CellMap<T>` that allows you to test a predicate on the internal value of a `Cell<T>`

Examples
```rust
// New
let cell = Cell::from("Hello World!".to_owned());
assert!(cell.map_(|s| s.contains("Hello"));
assert_eq!(cell.take(), "Hello World!");

// Old
let cell = Cell::from("Hello World!".to_owned());
let s = cell.take();
assert!(s.contains("Hello"));
cell.set(s);
assert_eq!(cell.take(), "Hello World!");
```

This also changes a few tests to use this interface, `test_condputs` and `test_number`

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -9,6 +9,7 @@ macro_rules! regex {
 }
 
 use crate::char_len;
+use std::cell::Cell;
 use std::{collections::HashMap, iter::FromIterator};
 
 pub struct TranslationTable {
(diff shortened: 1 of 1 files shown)


pickfire:
IMHO I don't think it is worth it, we could not utilize `assert_eq!` with this. Most of the cases, we could just `cell.take()` but we didn't since we want to keep the output, we could do that as well.

```
let cell = Cell::from("Hello World!".to_owned());
let s = cell.take();
assert!(s.contains("Hello"));
cell.set(s);
assert_eq!(cell.take(), "Hello World!");
```

It could be:
```
let cell = Cell::from("Hello World!".to_owned());
let s = cell.take();
assert!(s.contains("Hello"));
assert_eq!(s, "Hello World!");
assert_eq!(cell.take(), ""); // not tested
```

pickfire:
@charliethomson Can we close this? I don't see much benefit this helper provides, it helps only a bit but makes it not ergonomic (`_map`).

pickfire:
@charliethomson I am sorry about it. Thanks for the patch. :D

PR #94, opened 2020-06-16
feat: add parse_manpage_at_path

pickfire:


src/main.rs
```diff
@@ -1202,115 +1203,113 @@ fn cleanup_autogenerated_file(path: &Path) {
...
+    let cmdname = String::from_utf8_lossy(cmdname.unwrap());
+    let ignored_commands = [
+        "cc", "g++", "gcc", "c++", "cpp", "emacs", "gprof", "wget", "ld", "awk",
+    ];
+    if ignored_commands.contains(&cmdname.as_ref()) {
+        return Ok(false);
+    }
+
+    // Ignore perl's gazillion man pages
+    let ignored_prefixes = ["perl", "zsh"];
```
charliethomson, line 1252:
```rust
if ["perl", "zsh"]
    .iter()
    .any(|prefix| cmdname.starts_with(prefix)) 
{
    return Ok(false);
}

File: src/main.rs [hunk 2]
```diff
@@ -1202,115 +1203,113 @@ fn cleanup_autogenerated_file(path: &Path) {
...
+    {
+        let comments = format!(
+            "# {}\n#Autogenerated from man page {}\n",
+            &cmdname,
+            manpage_path.display()
+        );
+        completions.insert_str(0, &comments);
+
+        if write_to_stdout {
+            let stdout = io::stdout();
```
charliethomson, line 1289:
```rust
io::stdout()
    .lock()
    .write_all(completions.as_bytes())?;
```

src/main.rs
```diff
@@ -1202,115 +1203,113 @@ fn cleanup_autogenerated_file(path: &Path) {
...
+            let mut stdout = stdout.lock();
+            stdout.write_all(completions.as_bytes())?;
+        } else {
+            let fullpath = output_directory
+                .join(cmdname.as_ref())
+                .with_extension("fish");
+            let file = File::create(fullpath);
+            if file.is_err() {
```
charliethomson, line 1297:
```rust
match File::create(fullpath) {
    Ok(file) => file.write_all(completions.as_bytes())?,
    Err(e) => ..
}
```

src/main.rs
```diff
@@ -1202,115 +1203,113 @@ fn cleanup_autogenerated_file(path: &Path) {
...
+    }
+
+    // Get the "base" command, e.g. gcc.1.gz -> gcc
+    // These casts are safe as OsStr is internally a wrapper around [u8] on all
+    // platforms. Taken from libstd.
+    let cmdname = manpage_path
+        .file_name()
+        .and_then(|file| file.as_bytes().splitn(2, |b| *b == b'.').next());
+    let cmdname = String::from_utf8_lossy(cmdname.unwrap());
+    let ignored_commands = [
```
charliethomson, line 1244:
```rust
if [
    "cc", "g++", "gcc",
    "cpp", "emacs", "gprof",
    "wget", "ld", "awk"
].contains(&cmdname.as_ref()) {
    return Ok(false);
}
```

src/main.rs
```diff
@@ -1202,115 +1203,113 @@ fn cleanup_autogenerated_file(path: &Path) {
...
-//             if output_file != sys.stdout:
-//                 output_file.close()
-//         else:
-//             parser_names =  ', '.join(p.__class__.__name__ for p in parsersToTry)
-//             #add_diagnostic('%s contains no options or is unparsable' % manpage_path, BRIEF_VERBOSE)
-//             add_diagnostic('%s contains no options or is unparsable (tried parser %s)' % (manpage_path, parser_names), BRIEF_VERBOSE)
-//
-//     return success
+fn parse_manpage_at_path(
+    manpage_path: &Path,
```
charliethomson, line 1207:
```rust
fn parse_manpage_at_path<P: AsRef<Path>(
    manpage_path: P,
    output_directory: P,
    ..
}
```
?

pickfire: I don't get what you want to convey but I guess that you ask why is there additional fields not in the python version.

It is because some of the variables are available globally, so we need to make it locally by passing it through function.

src/main.rs
pickfire: I think the current one make it much easier to know those variables are part of ignored commands, the version you show does not indicate that.

src/main.rs
pickfire: I used that at first then I kept the original variable name back to make it easier to understand.

src/main.rs
pickfire: Nice, looks like a good option. Adding this.

src/main.rs
pickfire: Oh, good suggestion.

src/main.rs
charliethomson: i mean `&Path` is a more restrictive type than `AsRef<Path>` because `str`, `String`, `Path`, and `PathBuf` implement it.

src/main.rs
charliethomson: Fair

src/main.rs
pickfire: Yes, it is more restrictive. But we don't need to use generics here, we are not exposing a library, until we need it we could keep it as `&Path`. This could be the reason compilation time blow up, I think taking what we need to enough.

charliethomson: Fair

charliethomson:
LGTM

pickfire:
Just left @scooter-dangle to give his thumbs up.

I will be waiting until log put up their next release with the upcoming structured logging feature to add logging for fish-manpage-completions. I wonder how good it will go, in the meantime I can chill and work on oxidizing dwm (inspired by this repository).

PR #96, opened 2020-07-13
impl: parse_and_output_manpages

charliethomson:


Cargo.toml
```diff
@@ -15,4 +15,4 @@ bzip2 = "0.3"
 xz2 = "0.1"
 
 [dev-dependencies]
-pretty_assertions = "*"
+pretty_assertions = "*"
```
scooter-dangle, line 18: What editor are you using? You might want to modify its settings to ensure it doesn't strip the final line terminator.

scooter-dangle, line 18:
```suggestion
pretty_assertions = "*"

```

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+// TODO: Result<cmd_name, CompletionsError>
+fn parse_and_output_man_pages<P: AsRef<Path>>(
+    paths: impl Iterator<Item = P>,
+    output_directory: P,
+    Progress(show_progress): Progress,
+    deroff_only: bool,
+    write_to_stdout: bool,
+) -> Result<(), String> {
+    let mut paths = paths
+        .map(|path| path.as_ref().to_owned())
```
pickfire, line 1390: Is `to_owned` necessary? Can we just use the borrowed version?

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+        .collect::<Vec<_>>();
+    paths.sort();
+    paths.dedup();
+
+    let total_count = paths.len();
+
+    let (mut successful_count, mut index) = (0, 0);
+
+    let mut cmd_name = "".to_owned();
```
pickfire, line 1399: I think we should keep a TODO here.

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+
+    // Get the number of digits in total_count
+    let padding_len = (total_count as f32).log(10.) as usize + 1;
```
pickfire, line 1402: Err, what is this? Did you mean `total_count.to_string().len()`?

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
-    _output_directory: PathBuf,
-    Progress(_show_progress): Progress,
-) {
-    unimplemented!();
+// Output: Result<cmd_name, error>
+// TODO: Result<cmd_name, CompletionsError>
+fn parse_and_output_man_pages<P: AsRef<Path>>(
+    paths: impl Iterator<Item = P>,
+    output_directory: P,
+    Progress(show_progress): Progress,
```
pickfire, line 1385: I don't think we need `Progress`, we can just `bool`.

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+            .expect("Failed to get name of manpage path")
+            .to_string_lossy()
+            .to_mut()
+            .to_owned();
+
+        // gcc.1.gz -> gcc
+        cmd_name = match man_file_name.split('.').next() {
+            Some(cmd_name) => cmd_name.to_owned(),
+            None => return Err(format!("Failed to get cmd_name from {}", man_file_name)),
+        };
```
pickfire, line 1430: I wrote something similar in `parse_manpage_at_path`.
```rust
let cmdname = manpage_path
    .file_name()
    .and_then(|file| file.as_bytes().splitn(2, |b| *b == b'.').next());
let cmdname = String::from_utf8_lossy(cmdname.unwrap());
```

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+                    "{}{}",
+                    (0..(padding_len - cur_len)).fold("".to_owned(), |mut acc, _| {
+                        acc.push_str(" ");
+                        acc
+                    }),
+                    index
+                ),
+                total_count,
+                man_file_name,
+            );
```
pickfire, line 1448:
```suggestion
            let progress_str = format!(
                "  {:<1$} / {} : {}",
                index,
                padding_len,
                total_count,
                man_file_name,
            );
```
Rust does provide padding even though it is a bit hidden in the documentation.

Ideally, we should use some kind of progress bar that is more visually appealing than this later on.

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+
+        // gcc.1.gz -> gcc
+        cmd_name = match man_file_name.split('.').next() {
+            Some(cmd_name) => cmd_name.to_owned(),
+            None => return Err(format!("Failed to get cmd_name from {}", man_file_name)),
+        };
+
+        if show_progress {
+            // len(str(index))
+            let cur_len = (index as f32).log(10.) as usize + 1;
```
pickfire, line 1434: Not necessary.

src/main.rs
pickfire, line 1448:
```suggestion
            let progress_str = format!(
                "  {:>1$} / {} : {}",
                index,
                padding_len,
                total_count,
                man_file_name,
            );
```
Rust does provide padding in fmt, although it is a bit hidden in documentation.

Ideally, we will replace this afterwards with a more visually appealing progress bar.

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+                "{}{}",
+                progress_str,
+                (0..(last_progress_string_length - progress_str.len())).fold(
+                    "".to_owned(),
+                    |mut acc, _| {
+                        acc.push_str(" ");
+                        acc
+                    }
+                )
+            );
```
pickfire, line 1461:
```suggestion
            // padded_progress_str = progress_str.ljust(last_progress_string_length)
            let padded_progress_str = format!("{:<1$}", progress_str, last_progress_string_length);
```
We could throw out those `_str` and `_length` since we are using rust here. `progress` and `last_progress_length` should be fairly understandable.

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+
+            last_progress_string_length = progress_str.len();
+            // TODO: Expects
+            let stdout = std::io::stdout();
+            let mut lock = stdout.lock();
+            lock.write(format!("\r{}\r", padded_progress_str).as_bytes())
+                .expect("Failed to write to stdout");
+
+            lock.flush().expect("Failed to flush stdout");
```
pickfire, line 1470: Do we need flush? I thought it automatically flush when it goes out of scope? Is this good (not formatted)?
```rust
            io::stdout().write(format!("\r{}\r", padded_progress_str).as_bytes())
                .expect("Failed to write to stdout");
```

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+            write_to_stdout,
+        ) {
+            Ok(successful) => {
+                if successful {
+                    successful_count += 1
+                }
+                // Theres no else here, i assume they just pass over it lol
+            }
+            Err(e) => {
+                // add diag, io error
```
pickfire, line 1486:
```suggestion
                // add_diagnostic!(format!("Cannot open {}", manpage_path));
```
We can keep like the rest as of now so I can easy move to log later.

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+            deroff_only,
+            write_to_stdout,
+        ) {
+            Ok(successful) => {
+                if successful {
+                    successful_count += 1
+                }
+                // Theres no else here, i assume they just pass over it lol
+            }
+            Err(e) => {
```
pickfire, line 1485: I wonder if we should handle "cannot open {manpage_path}" like the original version? I guess we could skip interrupted error? Not sure about this.

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
+            }
+        };
+
+        // flush diagnostics
+    }
+
+    // "Newline after loop"
+    println!("");
+
+    // add diagnostic: successfully parsed {} / {} pages, successful_count, total_count
```
pickfire, line 1496:
```suggestion
    // add_diagnostic(format!("successfully parsed {} / {} pages", successful_count, total_count), BRIEF_VERBOSE);
```

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
-    unimplemented!();
+// Output: Result<cmd_name, error>
+// TODO: Result<cmd_name, CompletionsError>
+fn parse_and_output_man_pages<P: AsRef<Path>>(
+    paths: impl Iterator<Item = P>,
+    output_directory: P,
+    Progress(show_progress): Progress,
+    deroff_only: bool,
+    write_to_stdout: bool,
+) -> Result<(), String> {
```
pickfire, line 1388: Did we even return an error here? I believe there is no return from this. Like a one way ticket to heaven.

src/main.rs
```diff
@@ -1376,13 +1377,126 @@ fn parse_manpage_at_path(
...
-fn parse_and_output_man_pages(
-    _paths: impl Iterator<Item = PathBuf>,
-    _output_directory: PathBuf,
-    Progress(_show_progress): Progress,
-) {
-    unimplemented!();
+// Output: Result<cmd_name, error>
+// TODO: Result<cmd_name, CompletionsError>
+fn parse_and_output_man_pages<P: AsRef<Path>>(
+    paths: impl Iterator<Item = P>,
```
pickfire, line 1383: This could be simply `Vec<PathBuf>` or `Vec<Path>`.

src/main.rs
charliethomson: 1f0458c

src/main.rs
charliethomson: [This](https://github.com/rust-dc/fish-manpage-completions/commit/a5270bcd0e9a98b3da2d55aabff1689e67a443d1) was a @scooter-dangle thing

src/main.rs
charliethomson: da160a2?

src/main.rs
charliethomson: Ah, wasn't aware of that lol

src/main.rs
charliethomson: 57c0602

src/main.rs
charliethomson: 1f0458c

src/main.rs
charliethomson: I'm not sure. @scooter-dangle do you know?

pickfire: Let me ping @lonami, I wonder if he knows.

src/main.rs
pickfire: @charliethomson Is this incorrect?

pickfire: Not sure if log is faster, but you could do `log10()`.

src/main.rs
```diff
@@ -1311,78 +1312,119 @@ fn parse_manpage_at_path(
...
 struct Progress(pub bool);
 
-// TODO Arg/output types?
+// Output: Result<cmd_name, error>
+// TODO: Result<cmd_name, CompletionsError>
 fn parse_and_output_man_pages(
-    _paths: impl Iterator<Item = PathBuf>,
-    _output_directory: PathBuf,
-    Progress(_show_progress): Progress,
+    paths: Vec<PathBuf>,
```
pickfire, line 1321:
```suggestion
    paths: &[PathBuf],
```
It is recommended to use the one usable from coercion.

src/main.rs
```diff
@@ -1311,78 +1312,119 @@ fn parse_manpage_at_path(
...
+            Err(e) => {
+                // add_diagnostic(format!("Cannot open {}", manpage_path), VERY_VERBOSE)
+            }
+        };
+
+        // flush diagnostics
+    }
+
+    // "Newline after loop"
+    println!("");
```
pickfire, line 1424:
```suggestion
    println!();
```

src/main.rs
charliethomson: Log is O(1), `to_string().len()` is O(2N)
I didn't know log10 was a thing. I'll swap that real quick

src/main.rs
```diff
@@ -1311,78 +1312,119 @@ fn parse_manpage_at_path(
...
+                "Failed to get command name from {}",
+                man_file_name
+            ));
+
+        if show_progress {
+            let num_spaces = padding - num_digits(index);
+
+            let progress = format!(
+                "  {0:>1$} / {2} : {3}",
+                index, num_spaces, total, man_file_name,
```
pickfire, line 1383:
```suggestion
                index, padding, total, man_file_name,
```
padding already subtracts it for us.

src/main.rs
pickfire: Ah, I didn't know about that. This method does seemed interesting where do you find this? I tried `ceil`, that does not work, seemed like we need `+ 1`.

src/main.rs
```diff
@@ -1311,78 +1312,119 @@ fn parse_manpage_at_path(
...
+            let stdout = std::io::stdout();
+            let mut lock = stdout.lock();
+            lock.write(format!("\r{}\r", padded).as_bytes())
+                .expect("Failed to write to stdout");
+
+            lock.flush().expect("Failed to flush stdout");
+        }
+
+        match parse_manpage_at_path(
+            manpage_path.deref(),
```
pickfire, line 1404:
```suggestion
            &manpage_path,
```

src/main.rs
```diff
@@ -3,6 +3,7 @@ use std::collections::{HashMap, HashSet};
 use std::ffi::OsStr;
 use std::fs::{self, File};
 use std::io::{self, BufRead, BufReader, Read, Write};
+use std::ops::Deref;
```
pickfire, line 6: Probably not needed.

src/main.rs
```diff
@@ -1311,78 +1312,119 @@ fn parse_manpage_at_path(
...
+    Progress(show_progress): Progress,
+    deroff_only: bool,
+    write_to_stdout: bool,
 ) {
-    unimplemented!();
+    // Get the number of digits in num
+    #[inline(always)]
+    fn num_digits(num: usize) -> usize {
+        (num as f32).log10() as usize + 1
+    }
```
pickfire, line 1331: Closure would be simpler to do this job.
```suggestion
    let num_digits = |n| (n as f32).log10() as usize + 1;
```

src/main.rs
charliethomson: oops

src/main.rs
```diff
@@ -1311,78 +1312,119 @@ fn parse_manpage_at_path(
...
+            output_directory.as_ref(),
+            deroff_only,
+            write_to_stdout,
+        ) {
+            Ok(successful) => {
+                if successful {
+                    successful_count += 1
+                }
+                // Theres no else here, i assume they just pass over it lol
+            }
```
pickfire, line 1414:
```suggestion
            Ok(true) => successful_count += 1,
            Ok(false) => {}
```

src/main.rs
```diff
@@ -1311,78 +1312,119 @@ fn parse_manpage_at_path(
...
+            let progress = format!(
+                "  {0:>1$} / {2} : {3}",
+                index, num_spaces, total, man_file_name,
+            );
+
+            // padded_progress_str = progress_str.ljust(last_progress_string_length)
+            let padded = format!(
+                "{0:<1$}",
+                progress,
+                last_len.checked_sub(progress.len()).unwrap_or(0)
```
pickfire, line 1383: Do we need to subtract here?

By the way, what you are doing here is known as `saturating_sub`.

src/main.rs
charliethomson: :shrug: one of my friends is really good at that type of thing, she let me know about it a few months back. Pretty good solution tbh

src/main.rs
charliethomson: No, we don't.

src/main.rs
```diff
@@ -1405,12 +1399,8 @@ fn parse_and_output_man_pages(
...
             write_to_stdout,
         ) {
-            Ok(successful) => {
-                if successful {
-                    successful_count += 1
-                }
-                // Theres no else here, i assume they just pass over it lol
-            }
+            Ok(true) => successful_count += 1,
+            Ok(false) => (),
```
pickfire, line 1399: There's a clippy lint to change this to `{}` later, so we can just use `{}` rather than `(),`. I forgot which one is it already.

src/main.rs
Lonami: Not sure, the examples in the documentation for [`stdout`](https://doc.rust-lang.org/stable/std/io/fn.stdout.html) don't mention anything. Unless I'm looking in the wrong place perhaps the docs could be updated to note the behaviour. https://github.com/rust-lang/rust/issues/23205 makes me thing it'll be flushed.

src/main.rs
scooter-dangle: Especially when there are multiple arguments to a function of the same type and where order isn't obvious, it's a pattern for trivially ensuring the caller doesn't accidentally mix up the argument order. Not sure though why I initially added it when there was only one boolean arg to that function. :man_shrugging:

src/main.rs
```diff
@@ -1311,78 +1311,99 @@ fn parse_manpage_at_path(
...
-    Progress(_show_progress): Progress,
+    paths: &mut [PathBuf],
+    output_directory: PathBuf,
+    Progress(show_progress): Progress,
+    deroff_only: bool,
+    write_to_stdout: bool,
 ) {
-    unimplemented!();
+    // Get the number of digits in num
+    let num_digits = |n| (n as f32).log10() as usize + 1;
```
scooter-dangle, line 1327: Wouldn't hurt to promote this to a full function so we could add a couple test case assertions for the future reader.

src/main.rs
```diff
@@ -1311,78 +1311,99 @@ fn parse_manpage_at_path(
...
+
+    let mut paths = paths;
+    paths.sort();
+
+    let total = paths.len();
+
+    let (mut successful_count, mut index) = (0, 0);
```
scooter-dangle, line 1334: Kinda weird to construct a tuple that you immediately deconstruct.

```suggestion
    let mut successful_count = 0;
    let mut index = 0;
```

src/main.rs
scooter-dangle: I think the `flush` question is interesting in general, but in this case, is there any reason for us to not just `print!` or `println!` instead of explicitly manipulating `stdout`?

src/main.rs
```diff
@@ -1311,78 +1311,99 @@ fn parse_manpage_at_path(
...
+        println!(
+            "Parsing man pages and writing completions to {:?}",
+            output_directory
+        );
+    }
+
+    // As far as I can tell, we're fully supporting xz / lzma, so the checks here aren't needed
+
+    for manpage_path in paths {
+        index += 1;
```
scooter-dangle, line 1352: Not sure if the following is correct, but generally would be more idiomatic to obtain the index by using `Iterator::enumerate` than creating+incrementing a counter.
```suggestion
    for (index, manpage_path) in paths.iter().enumerate() {
```

src/main.rs
pickfire: The easiest way is to just borrow the whole options from clap arguments.

src/main.rs
```diff
@@ -1311,78 +1311,99 @@ fn parse_manpage_at_path(
...
+            .file_name()
+            .map(|fname| fname.to_string_lossy().to_mut().to_owned())
+            .unwrap_or_else(|| panic!("Failed to get manfile name from {:?}", manpage_path));
+
+        // gcc.1.gz -> gcc
+        cmd_name = man_file_name
+            .split('.')
+            .next()
+            .map(|cmd_name| cmd_name.to_owned())
+            .unwrap_or_else(|| panic!("Failed to get command name from {}", man_file_name));
```
pickfire, line 1365: Seemed to be doing work twice by having 2 strings, `man_file_name` could be `&str` or borrowed without creating a new `String`.

src/main.rs
pickfire:
```suggestion
    let num_digits = |n| (1.max(n) as f32).log10() as usize + 1;
```
Input 0 should be 1, not 0.

Validated by
```rust
fn main() {
    let num_digits = |n| (1.max(n) as f32).log10() as usize + 1;
    for i in 0..1000 {
        assert_eq!(num_digits(i), i.to_string().len(), "{}", i);
    }
}
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3d1e5e007938986c1211db3006ebcf72

src/main.rs
pickfire: Index is probably not needed.

src/main.rs
pickfire: Haha, you suggested the same thing I did (but I can't find it), except your index is taken from the wrong thing. This should be
```suggestion
    for (index, manpage_path) in (1..).zip(&paths) {
        index += 1;
```
`enumerate` starts from 0 but we want 1. This should be replaced by some indicator or progress bar later.

src/main.rs
charliethomson: Thanks Ivan

src/main.rs
charliethomson:
```rust
.map(|fname| fname.to_string_lossy())
```?

File: src/main.rs (hunk 5 above)
Comment (charliethomson) [reply src/main.rs]: It's removed. We can add it back if and when we add them for the rest of the bools

File: src/main.rs (hunk 28 above)
Comment (pickfire) [reply src/main.rs]: I meant `to_owned()`, `...to_string_lossy` becomes `Cow` so it may not reallocate. I don't know what `to_mut` is for though. First, `to_owned`. Second, `cmd_name.to_owned()`.

File: src/main.rs [hunk 29]
```diff
@@ -1311,78 +1311,108 @@ fn parse_manpage_at_path(
...
+        let man_file_name = manpage_path
+            .file_name()
+            .map(|fname| fname.to_string_lossy())
+            .unwrap_or_else(|| panic!("Failed to get manfile name from {:?}", manpage_path));
+
+        // gcc.1.gz -> gcc
+        cmd_name = man_file_name
+            .split('.')
+            .next()
+            .unwrap_or_else(|| panic!("Failed to get command name from {}", man_file_name));
```
pickfire, line 1371:
```suggestion
    let padding = num_digits(total);

    let mut last_len: usize = 0;

    if show_progress && !write_to_stdout {
        println!(
            "Parsing man pages and writing completions to {:?}",
            output_directory
        );
    }

    // As far as I can tell, we're fully supporting xz / lzma, so the checks here aren't needed

    for (index, manpage_path) in paths.iter().enumerate() {
        // foo/bar/gcc.1.gz -> gcc.1.gz
        // TODO: Expect
        let man_file_name = manpage_path
            .file_name()
            .map(|fname| fname.to_string_lossy())
            .unwrap_or_else(|| panic!("Failed to get manfile name from {:?}", manpage_path));

        // gcc.1.gz -> gcc
        let cmd_name = man_file_name
            .split('.')
            .next()
            .unwrap_or_else(|| panic!("Failed to get command name from {}", man_file_name));
```

charliethomson: If I understand, `let cmd_name = "";` => `let cmd_name = man_file_name...`?

pickfire: Kinda, we just move `cmd_name` around and reduce the scope.

src/main.rs
scooter-dangle: Seems like we do need to flush. Marking as resolved.

src/main.rs
```diff
@@ -1311,78 +1311,105 @@ fn parse_manpage_at_path(
...
+        let man_file_name = manpage_path
+            .file_name()
+            .map(|fname| fname.to_string_lossy())
+            .unwrap_or_else(|| panic!("Failed to get manfile name from {:?}", manpage_path));
+
+        // gcc.1.gz -> gcc
+        let cmd_name = man_file_name
+            .split('.')
+            .next()
+            .unwrap_or_else(|| panic!("Failed to get command name from {}", man_file_name));
```
scooter-dangle, line 1368: No need to have special panic message since this is an infallible `None`. Okay to just `unwrap` since `string.split('.').next().is_some()` is `true` for all possible strings. Maybe add a comment to that effect, e.g.:
```suggestion
            .unwrap(); // `str::split` iterator ALWAYS has a first element by definition
```

src/main.rs
```diff
@@ -1311,78 +1311,105 @@ fn parse_manpage_at_path(
...
+
+        if show_progress {
```
scooter-dangle, line 1370: Hmm. This is the same condition as in the original, but is it correct to ignore whether `write_to_stdout` is set to `true`? Could the progress printing end up mucking with the contents written to stdout? (Maybe that's why there's a `\r` before _and_ after the message we print—so that what we might write to stdout in `parse_manpage_at_path` ends up overwriting the progress bar on the terminal? That would work for terminal display but would still end up the progress messages onto the same stream as the actual content.

src/main.rs
```diff
@@ -1311,78 +1311,105 @@ fn parse_manpage_at_path(
...
-    _output_directory: PathBuf,
-    Progress(_show_progress): Progress,
+    paths: &mut [PathBuf],
+    output_directory: PathBuf,
+    show_progress: bool,
+    deroff_only: bool,
+    write_to_stdout: bool,
 ) {
-    unimplemented!();
+    let mut paths = paths;
```
scooter-dangle, line 1337: Already mutable. I don't think we need this line.
```suggestion
```

src/main.rs
charliethomson: Wasn't aware of that, thanks

src/main.rs
charliethomson: Yeah, I meant to get that added, just forgot. Thanks for noticing, fixing now

src/main.rs
```diff
@@ -1311,78 +1311,101 @@ fn parse_manpage_at_path(
...
+    for (index, manpage_path) in paths.iter().enumerate() {
+        // foo/bar/gcc.1.gz -> gcc.1.gz
+        // TODO: Expect
+        let man_file_name = manpage_path
+            .file_name()
+            .map(|fname| fname.to_string_lossy())
+            .unwrap_or_else(|| panic!("Failed to get manfile name from {:?}", manpage_path));
+
+        // gcc.1.gz -> gcc
+
```
scooter-dangle, line 1362:
```suggestion
```

src/main.rs
```diff
@@ -1311,78 +1311,100 @@ fn parse_manpage_at_path(
...
+                padding,
+                total,
+                man_file_name,
+            );
+
+            // padded_progress_str = progress_str.ljust(last_progress_string_length)
+            let padded = format!("{0:<1$}", progress, last_len);
+
+            last_len = progress.len();
+            // TODO: Expects
```
scooter-dangle, line 1378: Is this still something that you needed to get to, or is it removable?

src/main.rs
```diff
@@ -1311,78 +1311,100 @@ fn parse_manpage_at_path(
...
+        if show_progress && !write_to_stdout {
+            let progress = format!(
+                "  {0:>1$} / {2} : {3}",
+                index + 1,
+                padding,
+                total,
+                man_file_name,
+            );
+
+            // padded_progress_str = progress_str.ljust(last_progress_string_length)
```
scooter-dangle, line 1374:
```suggestion
```

src/main.rs
```diff
@@ -1311,78 +1311,100 @@ fn parse_manpage_at_path(
...
+            "Parsing man pages and writing completions to {:?}",
+            output_directory
+        );
+    }
+
+    // As far as I can tell, we're fully supporting xz / lzma, so the checks here aren't needed
+
+    for (index, manpage_path) in paths.iter().enumerate() {
+        // foo/bar/gcc.1.gz -> gcc.1.gz
+        // TODO: Expect
```
scooter-dangle, line 1355: Is this `TODO` still valid?

src/main.rs
```diff
@@ -1311,78 +1311,100 @@ fn parse_manpage_at_path(
...
+    let mut last_len: usize = 0;
+
+    if show_progress && !write_to_stdout {
+        println!(
+            "Parsing man pages and writing completions to {:?}",
+            output_directory
+        );
+    }
+
+    // As far as I can tell, we're fully supporting xz / lzma, so the checks here aren't needed
```
pickfire, line 1351:
```suggestion
    // As far as I can tell, we're fully supporting xz / lzma, so we don't check
```

src/main.rs
```diff
@@ -1311,78 +1311,100 @@ fn parse_manpage_at_path(
...
+
+            // padded_progress_str = progress_str.ljust(last_progress_string_length)
+            let padded = format!("{0:<1$}", progress, last_len);
+
+            last_len = progress.len();
+            // TODO: Expects
+            let stdout = std::io::stdout();
+            let mut lock = stdout.lock();
+            lock.write_all(format!("\r{}\r", padded).as_bytes())
+                .expect("Failed to write to stdout");
```
pickfire, line 1382: I don't think we need to be so old rigid. Why not just clear current line so we don't need to keep track of how many items we have printed? `\e[K{}\r`, `\e[K` clear current line from cursor to end, we could remove `last_len` if we do this.

charliethomson: I'm not familiar with `\e[K\r`?

pickfire: Just change that to `\e[K{}\r`. It should work, I tested. Try on your terminal?
```
$ echo \e '\e[Khello world\r\e[Kyay\r'
```
It should show `yay`.

scooter-dangle: Looks like Rust isn't familiar with `\e` as an escape sequence.

charliethomson: @pickfire [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6d1d2e2c995894962aae994f4c81bc9e)

pickfire: Ah, you need to unescape that in rust, either use `r"\e[K{}\r"` or `"\\e[K{}\\r"`, I prefer the first.

charliethomson: Am I missing something?
```rust
// Contents of test.rs
fn main() { 
    for i in 0..10 {
        println!(r"\e[K{}\r", i);
            
    }
}
```
```
c@flesh ~/g/fish-manpage-completions (deroff)> nv test.rs
c@flesh ~/g/fish-manpage-completions (deroff)> rustc test.rs 
c@flesh ~/g/fish-manpage-completions (deroff)> ./test
\e[K0\r
\e[K1\r
\e[K2\r
\e[K3\r
\e[K4\r
\e[K5\r
\e[K6\r
\e[K7\r
\e[K8\r
\e[K9\r
```

charliethomson: We could just `// TODO: pickfire thinks this is ugly` this

pickfire:
```rust
> cat > src/main.rs
use std::io::{self, Write};
use std::thread;
use std::time::Duration;

fn main() {
    for i in 0..10 {
        let stdout = io::stdout();
        let mut stdout = stdout.lock();
        stdout.write_all(format!("\r\x1b[K{}", i).as_bytes()).ok();
        stdout.flush().ok();
        thread::sleep(Duration::from_secs(1));
    }
    println!();
}
> rustc main.rs
> ./main.rs
9
```

charliethomson: 10a0120

src/main.rs
```diff
@@ -1311,78 +1311,95 @@ fn parse_manpage_at_path(
...
+            );
+
+            let padded = format!("{0:<1$}", progress, last_len);
+
+            last_len = progress.len();
+            let stdout = std::io::stdout();
+            let mut lock = stdout.lock();
+            lock.write_all(format!("\r\x1b[K{}", padded).as_bytes())
+                .expect("Failed to write to stdout");
+
```
pickfire, line 1378:
```suggestion
```

src/main.rs
```diff
@@ -1311,78 +1311,95 @@ fn parse_manpage_at_path(
...
+                total,
+                man_file_name,
+            );
+
+            let padded = format!("{0:<1$}", progress, last_len);
+
+            last_len = progress.len();
+            let stdout = std::io::stdout();
+            let mut lock = stdout.lock();
+            lock.write_all(format!("\r\x1b[K{}", padded).as_bytes())
```
pickfire, line 1376: Are you sure we need to pad? Why not just use `progress` here?

src/main.rs
```diff
@@ -1311,78 +1311,95 @@ fn parse_manpage_at_path(
...
+            output_directory
+        );
+    }
+
+    for (index, manpage_path) in paths.iter().enumerate() {
+        // foo/bar/gcc.1.gz -> gcc.1.gz
+        let man_file_name = manpage_path
+            .file_name()
+            .map(|fname| fname.to_string_lossy())
+            .unwrap_or_else(|| panic!("Failed to get manfile name from {:?}", manpage_path));
```
pickfire, line 1356:
```suggestion
            .unwrap_or_else(|| panic!("Failed to get manfile name from {}", manpage_path.as_ref()));
```

src/main.rs
```diff
@@ -1311,78 +1311,95 @@ fn parse_manpage_at_path(
...
+
+    let mut successful_count = 0;
+    let padding = num_digits(total);
+
+    let mut last_len: usize = 0;
+
+    if show_progress && !write_to_stdout {
+        println!(
+            "Parsing man pages and writing completions to {:?}",
+            output_directory
```
pickfire, line 1347:
```suggestion
            "Parsing man pages and writing completions to {}",
            output_directory.as_ref()
```

src/main.rs
```diff
@@ -1311,78 +1311,95 @@ fn parse_manpage_at_path(
...
+
+        // gcc.1.gz -> gcc
+        // `str::split` iterator ALWAYS has a first element by definition
+        let cmd_name = man_file_name.split('.').next().unwrap();
+
+        if show_progress && !write_to_stdout {
+            let progress = format!(
+                "  {0:>1$} / {2} : {3}",
+                index + 1,
+                padding,
```
pickfire, line 1366: This should be called `max_digits` to be clear, not exactly padding since it includes the original space itself.

src/main.rs
charliethomson: AFAIK the original did it to overwrite the last line with whitespace before printing again

src/main.rs
charliethomson: These don't work, it can't tell what you're trying to `as_ref` as, also, i dont know why as_ref() is better in this context?

pickfire: Because it will be printed out as string.

src/main.rs
pickfire: This already does that
```rust
use std::io::{self, Write};
use std::thread;
use std::time::Duration;

fn main() {
    for i in (0..103).rev() {
        let stdout = io::stdout();
        let mut stdout = stdout.lock();
        stdout.write_all(format!("\r\x1b[K{}", if i == 100 { 9999999 } else { i }).as_bytes()).ok();
        stdout.flush().ok();
        thread::sleep(Duration::from_secs(1));
    }
    println!();
}
```

charliethomson: Ah, ok, I'll add that tonight

src/main.rs
charliethomson: Do you mean like `"path/to/whatever.zip"` -> `path/to/whatever.zip`?
if so, we'd need to do `manpage_path.as_path().display()`, because it can't tell if you're trying to `as_ref()` into an `OsStr` or a `Path`

pickfire: IIRC `as_ref()` will turn it into something that is `Display`, I don't know if it's an `OsStr` or `Path` too, I just know that it becomes something displayable.

src/main.rs
charliethomson: 69bca77

src/main.rs
charliethomson: PathBuf is `Deref<Target=Path>`, `Path` has a [`display`](https://doc.rust-lang.org/std/path/struct.Path.html#method.display) method that makes it `Display`, I'll add this as `manpage_path.display()`

charliethomson: 5cae975

src/main.rs
charliethomson: 5cae975

src/main.rs
```diff
@@ -1311,78 +1311,94 @@ fn parse_manpage_at_path(
...
+        // foo/bar/gcc.1.gz -> gcc.1.gz
+        let man_file_name = manpage_path
+            .file_name()
+            .map(|fname| fname.to_string_lossy())
+            .unwrap_or_else(|| {
+                panic!(
+                    "Failed to get manfile name from {:?}",
+                    manpage_path.display()
+                )
+            });
```
pickfire, line 1359: This seemed quite long for just one task, is clippy formatting this? Or can it be shortened?

pickfire: Can we put `panic!(...)` on the same line? Different line from `unwrap_or_else`.

charliethomson: thats a `fmt` thing

pickfire: Yes, but other than this it looks good.

charliethomson: Fmt automatically puts it on a separate line, and Travis runs fmt on the code and expects no changes, also it's not like it affects the legibility of the code either way

pickfire: rustfmt allows the user to format macros in some ways, I know the version you put is allowed, the one I suggested is also allowed (I tested it locally on new file). I think it is more readable with it being shorter when panic is on the same line if possible, like in this case.

pickfire:
@charliethomson You are correct, the panic I suggested was incorrect, I didn't test it correct. Sorry about the bad suggestion, can you please revert it? Once it is reverted I can merge this.

charliethomson:
@pickfire

PR #99, opened 2020-07-16
impl: deroff, to_tbl, spec

charliethomson:


src/deroff.rs
```diff
@@ -1028,19 +1044,107 @@ impl Deroffer {
...
+                    } else {
+                        let mut option = self.s.to_owned();
+                        let mut arg = String::new();
+
+                        let mut idx: usize = 0;
+
+                        while option
+                            .get(idx..=idx)
+                            .and_then(|s| Some(s.chars().all(|c| c.is_alphabetic())))
+                            .unwrap_or_default()
```
pickfire, line 1064:
```suggestion
                            .map_or(false, |s| s.chars().all(|c| c.is_alphabetic()))
```
Also, I think it might not be a good idea to `get()` with bytes when we are already iterating at a character level. If we want to do `get` we probably doesn't even need to think about `chars()` and use `bytes()` instead.

src/deroff.rs
```diff
@@ -1028,19 +1044,107 @@ impl Deroffer {
...
+    fn do_tbl(&mut self) -> bool {
+        match self.tblstate {
+            TblState::Options => {
+                while !self.s.is_empty() && !"\n;".contains(self.str_at(0)) {
+                    self.skip_leading_whitespace();
+
+                    if !self.str_at(0).chars().all(|c| c.is_alphabetic()) {
+                        self.skip_char(1);
+                    } else {
+                        let mut option = self.s.to_owned();
```
pickfire, line 1056: Just wondering, can we use `&str` for option? I think probably not since we are mutating `self.s` below.

src/deroff.rs
```diff
@@ -1028,19 +1044,107 @@ impl Deroffer {
...
+                        if option.get(idx..=idx) == Some("(") {
+                            option = option.get(..idx).unwrap_or_default().to_owned();
+                            self.s = self.s.get(idx + 1..).unwrap_or_default().to_owned();
+                            arg = self.s.clone();
+                        } else {
+                            self.s = "".to_owned();
+                        }
+
+                        if !arg.is_empty() {
+                            if let Some(idx_) = arg.find("(") {
```
pickfire, line 1078: In the python code, `idx` is mutated to `-1` if it is not found, whereas we do not.
```python
idx = arg.find(")")
if idx != -1:
    arg = arg[:idx]
self.s = self.s[idx + 1 :]
```

src/deroff.rs
```diff
@@ -1028,19 +1044,107 @@ impl Deroffer {
...
+                                idx = idx_;
+                                arg = arg.get(..idx).unwrap_or_default().to_owned();
+                            }
+                            self.s = self.s.get(idx + 1..).unwrap_or_default().to_owned();
```
pickfire, line 1082: In the python code, `idx` is mutated to `-1` if it is not found, whereas we do not.
```python
idx = arg.find(")")
if idx != -1:
    arg = arg[:idx]
self.s = self.s[idx + 1 :]
```
For the last line, our code may behave different when it is not found.
```suggestion
                            if let Some(idx) = arg.find("(") {
                                arg = arg.get(..idx).unwrap_or_default().to_owned();
                                self.s = self.s.get(idx + 1..).unwrap_or_default().to_owned();
                            }
```

src/deroff.rs
```diff
@@ -1028,19 +1044,107 @@ impl Deroffer {
...
+                        {
+                            idx += 1;
+                        }
+
+                        if option.get(idx..=idx) == Some("(") {
+                            option = option.get(..idx).unwrap_or_default().to_owned();
+                            self.s = self.s.get(idx + 1..).unwrap_or_default().to_owned();
+                            arg = self.s.clone();
+                        } else {
+                            self.s = "".to_owned();
```
pickfire, line 1074: I wonder if this would be better.
```suggestion
                            self.s.clear();
```

src/deroff.rs
```diff
@@ -1028,19 +1044,107 @@ impl Deroffer {
...
+                self.text();
+            }
+        }
+
+        true
+    }
+
+    fn do_line(&mut self) -> bool {
+        match self.s.chars().nth(0) {
+            Some('.') | Some('\'') => !self.request_or_macro(),
```
pickfire, line 1123:
```suggestion
        match self.s.bytes().next() {
            Some(b'.') | Some(b'\'') => !self.request_or_macro(),
```

src/deroff.rs
```diff
@@ -1028,19 +1044,107 @@ impl Deroffer {
...
+            Some(_) => {
                 if self.tbl {
-                    do_tbl(s)
+                    self.do_tbl()
                 } else {
-                    text(s)
+                    self.text()
                 }
             }
             None => panic!("do_line` called with empty string as argument"),
```
pickfire, line 1131:
```suggestion
            None => unreachable!(),
```
Or better, we could just use `expect` or `unwrap` since the original code does not handle this. So no need to match `Some`.

src/deroff.rs
```diff
@@ -1028,19 +1044,107 @@ impl Deroffer {
...
+        for mut line in lines {
+            line.push('\n');
+            self.s = line;
+            if !self.do_line() {
+                return false;
+            }
+        }
+
+        true
+    }
```
pickfire, line 1147:
```suggestion
    pub fn deroff(&mut self, s: String) {
        let lines = s.split('\n');
        for line in lines {
            self.s = line.to_owned() + "\n";
            if !self.do_line() {
                break
            }
        }
    }
```
I wish it could `\n`, hopefully someone can add that to standard library.

src/deroff.rs
```diff
@@ -1060,23 +1164,56 @@ fn test_comment() {
     assert_eq!(deroffer.s, "\nworld".to_owned());
 }
 
-fn deroff_files(files: &[String]) -> std::io::Result<()> {
+fn deroff_files<P: AsRef<std::path::Path>, S: ToString>(
+    files: &[S],
```
pickfire, line 1168: I think using `&[String]` or `&[&str]` is fine for me. If `ToString` is really needed, we could use `Display` but I think the original is better.

src/deroff.rs
```diff
@@ -1060,23 +1164,56 @@ fn test_comment() {
...
+    output_dir: P,
+) -> std::io::Result<()> {
+    let output_path = output_dir.as_ref().to_path_buf();
     for arg in files {
+        let arg = arg.to_string();
```
pickfire, line 1173: Not necessary if we use `&[String]`.

src/deroff.rs
```diff
@@ -1060,23 +1164,56 @@ fn test_comment() {
     assert_eq!(deroffer.s, "\nworld".to_owned());
 }
 
-fn deroff_files(files: &[String]) -> std::io::Result<()> {
+fn deroff_files<P: AsRef<std::path::Path>, S: ToString>(
```
pickfire, line 1167: We would need another binary for this, looks like we were missing this. @scooter-dangle 

To do this, we have to create `src/bin/deroff.rs`.

src/deroff.rs
```diff
@@ -1060,23 +1164,56 @@ fn test_comment() {
...
+                    // TODO: This is a _bad_ workaround for latin1, we need to correctly decode input files
+                    let mut bytes = Vec::new();
+                    file.read_to_end(&mut bytes)?;
+
+                    for byte in bytes {
+                        string.push(byte as char);
+                    }
+                }
+                _ => {}
+            };
```
pickfire, line 1193: Is this even necessary? I wonder if we could drop this feature altogether, who uses `latin1` nowdays? UTF-8 for the win.

src/deroff.rs
```diff
@@ -1060,23 +1164,56 @@ fn test_comment() {
...
         }
         let mut d = Deroffer::new();
         println!("string: {}", string);
 
         d.deroff(string);
-        d.flush_output(std::io::stdout());
```
pickfire, line 1079: I thought this is correct?
```python
def deroff_files(files):
    for arg in files:
        sys.stderr.write(arg + "\n")
        if arg.endswith(".gz"):
            f = gzip.open(arg, "r")
            str = f.read()
            if IS_PY3:
                str = str.decode("latin-1")
        else:
            f = open(arg, "r")
            str = f.read()
        d = Deroffer()
        d.deroff(str)
        d.flush_output(sys.stdout)
        f.close()
```
It flushes to `sys.stdout` rather than file.

src/deroff.rs
charliethomson: I forgot about `.get()` getting bytes not chars. Lemme fix that real quick

src/deroff.rs
charliethomson:
```rust
match self.s.bytes().next().unwrap_or_else(|| unreachable!()) { .. }
```

src/deroff.rs
charliethomson: You mean `impl<S: ToString> Add<S> for String { .. }` ?

src/deroff.rs
charliethomson: I like your idea of putting this and things like it in it's own container, I'll make it faithful to the original and comment it out with a TODO

src/deroff.rs
charliethomson: The original explicitly supports python, I think we should outdo them by supporting more encodings. I'm thinking about another module that will take a byte stream in any encoding and convert it to a utf-8 string, thoughts?

src/deroff.rs
charliethomson: Yeah, we need to clone sadly

src/deroff.rs
charliethomson: I dont know why it sent two different messages there but basically
the python supports latin1, we should do more, maybe a thing in util or something (post-mvp)

scooter-dangle: > Is this even necessary? I wonder if we could drop this feature altogether, who uses `latin1` nowdays? UTF-8 for the win.

Regarding `latin1` support, we have a couple samples in rust-dc/manpages that use `latin1`. For MVP, wouldn't want to strip functionality compared to the original unless the original functionality is inherently buggy or insecure.

src/deroff.rs
scooter-dangle: Either that or add an option to the top-level binary that just deroffs files. Probably makes sense to have a binary that just deroffs?

Either way, will make it easier for us to test the deroffing against the contents of https://github.com/rust-dc/manpages.

charliethomson: @scooter-dangle @pickfire I'll work on this tonight and make a pr

pickfire: Interesting, I didn't know @kbknapp thought about this so fast. He even have a test repository prepared.

Ideally, not a top-level binary, just a binary for deroff like the original python `deroff.py`, I don't think it is useful for users to run this.

@charliethomson Take your time, we can fight the final boss slower. I wonder if would be cool if everyone can fight it together, like in a meetup.

src/deroff.rs
pickfire: https://github.com/rust-dc/manpages is latin only, I wonder what will happen if we add non-latin manpages into the mix. We throw everything from my /usr/share/man there. I think I have like 30609 man pages on my machine, sometimes I do try to read chinese manpages for fun. We could also use it for benchmark later on other than just fuzzing.

src/deroff.rs
```diff
@@ -1028,17 +1044,98 @@ impl Deroffer {
...
+                                self.s.drain(..=idx);
+                            }
+                        } else {
+                            // This was commented out in the python, so it is here too :)
+                            // self.skip_char(1);
+                        }
+
+                        if option.to_lowercase() == "tab" {
+                            self.tblTab = arg.get(0..1).unwrap_or_default().to_owned();
+                        }
```
pickfire, line 1088: Let me bend this a bit for you, needs some review. Heh, this one does no allocation.
```suggestion
                        // find first non-alphabetic character
                        match self.s.chars().position(|c| !c.is_alphabetic()) {
                            Some(idx) if self.s.get(idx..=idx) == Some("(") => {
                                // self.s -> option '(' arg '(' rest
                                let option = self.s.split_at(idx).0;
                                let mut iter = self.s.split_at(idx + 1).1.splitn(2, '(');
                                let arg = iter.next().unwrap_or_default();
                                let rest = iter.next().unwrap_or_default();

                                if option.to_lowercase() == "tab" {
                                    self.tblTab = arg.get(0..1).unwrap_or_default().to_owned();
                                }

                                // Likely faster alternative to self.s = rest
                                self.s.drain(self.s.len() - rest.len()..);
                            }
                            _ => self.s.clear(),
                        }
```

src/deroff.rs
```diff
@@ -1028,17 +1044,98 @@ impl Deroffer {
...
-            Some(c) => {
+    fn do_tbl(&mut self) -> bool {
+        match self.tblstate {
+            TblState::Options => {
+                while !self.s.is_empty() && !"\n;".contains(self.str_at(0)) {
+                    self.skip_leading_whitespace();
+
+                    if !self.str_at(0).chars().all(|c| c.is_alphabetic()) {
+                        self.skip_char(1);
+                    } else {
```
pickfire, line 1055: We should keep the original comment `Parse option` here.

src/deroff.rs
scooter-dangle: > https://github.com/rust-dc/manpages is latin only

Ah, yeah, I meant that there are only a few examples (that I knew of, at least) that were using characters in the Latin-1 non-ASCII range. Everything else is only trivially `latin1`.

> I wonder what will happen if we add non-latin manpages into the mix. We throw everything from my /usr/share/man there. I think I have like 30609 man pages on my machine, sometimes I do try to read chinese manpages for fun. We could also use it for benchmark later on other than just fuzzing.

For sure. It would make sense to expand to other encodings.

scooter-dangle:
> `do_tbl` and `spec` can be easily tested, I think we should add tests for them.

Seconded. If we're at a point where we don't have to worry about `unreachable!` code, should have tests. For example, they'd simplify verifying @pickfire's reduced-allocation refactor.

src/deroff.rs
```diff
@@ -1028,17 +1044,83 @@ impl Deroffer {
...
```
pickfire, line 1055:
```suggestion
                    } else { // parse option
```

src/deroff.rs
```diff
@@ -1028,17 +1044,83 @@ impl Deroffer {
...
+                        // find first non-alphabetic character
+                        match self.s.chars().position(|c| !c.is_alphabetic()) {
+                            Some(idx) if self.s.get(idx..=idx) == Some("(") => {
+                                // self.s -> option '(' arg '(' rest
+                                let option = self.s.split_at(idx).0;
```
scooter-dangle, line 1060:
```suggestion
                                let option = self.s[..idx];
```

src/deroff.rs
```diff
@@ -1028,17 +1044,83 @@ impl Deroffer {
...
+                                let mut iter = self.s.split_at(idx + 1).1.splitn(2, '(');
```
scooter-dangle, line 1061:
```suggestion
                                let mut iter = self.s[idx + 1..].splitn(2, '(');
```

src/deroff.rs
```diff
@@ -1028,17 +1044,83 @@ impl Deroffer {
...
+
+                self.text();
+            }
+        }
+
+        true
+    }
+
+    fn do_line(&mut self) -> bool {
+        match self.s.bytes().next().unwrap_or_else(|| unreachable!()) {
```
scooter-dangle, line 1105: Why do we claim the `None` case is unreachable here? I usually only reserve `unreachable!` for cases where I can locally verify it or where there is something inherent in a type's guarantees that isn't already captured by the type system.

Just a panic (via `unwrap` or `expect`) makes more sense here. E.g.,
```suggestion
        match self.s.bytes().next().expect("`do_line` called when `self.s` was empty") {
```

src/deroff.rs
```diff
@@ -1028,17 +1044,83 @@ impl Deroffer {
...
+            TblState::Options => {
+                while !self.s.is_empty() && !"\n;".contains(self.str_at(0)) {
+                    self.skip_leading_whitespace();
+
+                    if !self.str_at(0).chars().all(|c| c.is_alphabetic()) {
+                        self.skip_char(1);
+                    } else {
+                        // find first non-alphabetic character
+                        match self.s.chars().position(|c| !c.is_alphabetic()) {
+                            Some(idx) if self.s.get(idx..=idx) == Some("(") => {
```
scooter-dangle, line 1058: Can avoid having to re-lookup the character by using `char_indices`:
```suggestion
                        match self.s.char_indices().find(|(_, c)| !c.is_alphabetic()) {
                            Some((idx, ')')) => {
```

src/deroff.rs
pickfire: This won't compile if you combined the two together.

pickfire, line 1061:
```suggestion
                                let option = &self.s[..idx];
                                let mut iter = self.s[idx + 1..].splitn(2, '(');
```
suggested by @scooter-dangle

src/deroff.rs
pickfire: Because it really is unreachable.
https://github.com/rust-dc/fish-manpage-completions/pull/99/files#diff-5b0804df3203e0b4fc4f245618d61d53R1120

```rust
             self.s = line.to_owned() + "\n";
```
There will be at least `\n`.

src/deroff.rs
pickfire: TIL

src/deroff.rs
scooter-dangle: But yeah, that's what I mean. You have to rely on something that's not local to the function and not inherent to the type. If I as a dev at some point use this same function elsewhere in the module, there's nothing explicit to say that it's not meant to be called when `self.s` is non-empty.

src/deroff.rs
charliethomson: Are we sure we want to use raw indexing?

scooter-dangle: `&string[idx..]` is logically equivalent to `string.split_at(idx).1` (and same for `..idx` case). Same panic conditions. Using `split_at` when we're only grabbing one element of the resulting tuple feels like an odd avoidance of direct indexing.

charliethomson: cool, just wanted to make sure

src/deroff.rs
```diff
@@ -1058,18 +1058,17 @@ impl Deroffer {
                         // find first non-alphabetic character
                         match self.s.char_indices().find(|(_, c)| !c.is_alphabetic()) {
                             Some((idx, '(')) => {
-                                // self.s -> option '(' arg '(' rest
+                                // self.s -> option '(' arg ')' rest
                                 let option = &self.s[..idx];
-                                let mut iter = self.s[idx + 1..].splitn(2, '(');
+                                let mut iter = self.s[idx + 1..].splitn(2, ')');
```
pickfire, line 1063: Nice findings. My mistake.

PR #100, opened 2020-07-21
patch: FIx a few bugs in do_tbl, finish test_do_tbl

charliethomson:


src/deroff.rs
```diff
@@ -1216,18 +1222,68 @@ fn test_do_tbl() {
     assert!(deroffer.do_tbl());
     assert_eq!(deroffer.tblTab, "a");
     assert_eq!(deroffer.s, ";");
+    assert_eq!(deroffer.tblstate, TblState::Format);
 
     // </Options>
 
+    // <Format>
```
pickfire, line 1229: Haha, html tags.

src/deroff.rs
```diff
@@ -1182,27 +1184,31 @@ fn test_do_tbl() {
     assert!(deroffer.tblTab.is_empty());
     assert_eq!(deroffer.s, ";Hello World!");
     assert_eq!(deroffer.output.take(), "\n");
+    assert_eq!(deroffer.tblstate, TblState::Format);
```
pickfire, line 1187: I thought this was missing from the last tests and it indeed is.

src/deroff.rs
```diff
@@ -1216,18 +1222,68 @@ fn test_do_tbl() {
...
-    deroffer.s = "".into();
+    deroffer.s = "HelloaWorld!".into();
+    assert!(deroffer.do_tbl());
+    // This also calls text, so this checks that works correctly
+    // "Hello\tWorld!" passed to self.text
+    assert_eq!(deroffer.output.take(), "Hello\tWorld!");
+
+    // </Data>
+
+    // TODO: Add tests for the rest of the functions
```
pickfire, line 1286: Isn't this finished?

charliethomson: This test is finished, but there are a bunch of functions without tests (text being one of them)

charliethomson:
I broke it hold up

charliethomson:
It's a lot of commits but i dont want it to break again :disappointed:

src/deroff.rs
```diff
@@ -1001,6 +1002,8 @@ impl Deroffer {
     }
 
     fn flush_output<W: std::io::Write>(&mut self, mut write: W) {
+        println!("{}", self.get_output());
```
pickfire, line 1005: Oh, is this needed? Shouldn't we write to writer instead of stdout?

charliethomson: Oops that was a debugging thing  from a WHILE ago. fixed

src/deroff.rs
```diff
@@ -1092,7 +1095,7 @@ impl Deroffer {
             }
             TblState::Data => {
                 if !self.tblTab.is_empty() {
-                    self.s = self.s.replace(&self.tblTab, "\n");
+                    self.s = self.s.replace(&self.tblTab, "\t");
```
pickfire, line 1098: Oh, I did missed this. T_T

charliethomson: I'm really glad i made that test based on the python lol

src/deroff.rs
```diff
@@ -1157,7 +1159,18 @@ fn deroff_files(files: &[String]) -> io::Result<()> {
...
+                Err(_) => {
+                    // TODO: This is a _bad_ workaround for latin1, we need to correctly decode input files
+                    let mut bytes = Vec::new();
+                    file.read_to_end(&mut bytes)?;
+
+                    for byte in bytes {
+                        string.push(byte as char);
+                    }
+                }
+                _ => {}
```
pickfire, line 1172:
```suggestion
                Err(e) if e.kind() == io::ErrorKind::InvalidData => { // not valid UTF-8
                    // TODO: This is a _bad_ workaround for latin1, we need to correctly decode input files
                    let mut bytes = Vec::new();
                    file.read_to_end(&mut bytes)?;

                    for byte in bytes {
                        string.push(byte as char);
                    }
                }
                Err(e) => {
                    eprintln!("unknown error {:?}", e);
                    continue;
                }
                Ok(_) => {}
```
I think we should only handle `InvalidData` here. I think we should import `ErrorKind` and remove `io::`. Also, it would be easier if we work on byte level so we don't need to care if it's valid utf-8 or not.

src/deroff.rs
pickfire: I think we can leave that to a separate patch. This should be good enough to cover `do_tbl`.

src/deroff.rs
```diff
@@ -1216,18 +1238,68 @@ fn test_do_tbl() {
...
     let mut deroffer = Deroffer::new();
     deroffer.tblstate = TblState::Data;
+    deroffer.tblTab = "a".into();
 
-    deroffer.s = "".into();
+    deroffer.s = "HelloaWorld!".into();
+    assert!(deroffer.do_tbl());
+    // This also calls text, so this checks that works correctly
+    // "Hello\tWorld!" passed to self.text
+    assert_eq!(deroffer.output.take(), "Hello\tWorld!");
```
pickfire, line 1299:
```suggestion
    assert_eq!(deroffer.tblstate, TblState::Data);
    assert_eq!(deroffer.s, "");
    assert_eq!(deroffer.output.take(), "Hello\tWorld!");
```

src/deroff.rs
```diff
@@ -1182,27 +1201,31 @@ fn test_do_tbl() {
     assert!(deroffer.tblTab.is_empty());
     assert_eq!(deroffer.s, ";Hello World!");
     assert_eq!(deroffer.output.take(), "\n");
+    assert_eq!(deroffer.tblstate, TblState::Format);
```
pickfire, line 1204: I think we should arrange all of these in an order to make it easier to read. Or even better, use a macro. But I think this is sufficient.

charliethomson: I'm converting them to a macro right now :)

charliethomson: macro version is uglier, we're gonna keep it how it is :smile:

PR #101, opened 2020-07-21
impl: Tests for deroff.rs

charliethomson:
I implemented a test for every (non macro) function.
I'm still working on `test_deroff`, that may be a different PR, haven't decided yet, it just seems like a bigger fish. I noticed a lot of bugs while I was making the tests, so I fixed all of them and the behaviour matches that of the python source. If you guys notice anything that seems wrong definitely let me know.


charliethomson:
Switching to a branch with not 50 things on it

PR #104, opened 2020-08-02
Impl: rough system for handling non utf-8 input

charliethomson:
I'm open to suggestions, this "works", but it's not perfect (not sure a perfect solution exists, but yeah)

### DIFF
--- a/src/deroff.rs
+++ b/src/deroff.rs
@@ -1145,25 +1145,21 @@ fn deroff_files(files: &[String]) -> io::Result<()> {
             let mut decoder = GzDecoder::new(file);
             decoder.read_to_string(&mut string)?;
         } else {
-            match file.read_to_string(&mut string) {
-                Err(e) if e.kind() == io::ErrorKind::InvalidData => {
-                    // not valid UTF-8
-                    // TODO: This is a _bad_ workaround for latin1, we need to correctly decode input files
-                    let mut bytes = Vec::new();
-                    file.read_to_end(&mut bytes)?;
+            let mut bytes = Vec::new();
+            file.read_to_end(&mut bytes)?;
 
-                    for byte in bytes {
-                        string.push(byte as char);
+            match String::from_utf8(bytes.clone()) {
+                Ok(s) => string = s,
+                Err(e) => {
+                    // If UTF-8 Decoding failed, pass the work to encoding_rs_io
+                    println!("Unknown encoding, trying to decode");
+                    match crate::util::decode_bytes(bytes) {
+                        Some(s) => string = s,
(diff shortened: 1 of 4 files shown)


src/deroff.rs
```diff
@@ -1145,25 +1145,21 @@ fn deroff_files(files: &[String]) -> io::Result<()> {
...
-                    file.read_to_end(&mut bytes)?;
-
-                    for byte in bytes {
-                        string.push(byte as char);
-                    }
-                }
+            let mut bytes = Vec::new();
+            file.read_to_end(&mut bytes)?;
+
+            match String::from_utf8(bytes.clone()) {
```
pickfire, line 1151: I don't think cloning here is a good idea, I believe that broken encodings are kinda a rare case nowadays. I personally think ignoring broken utf-8 files is the better choice or maybe just use `String::from_utf8_lossy` to replace any broken utf-8 characters.

src/util.rs
```diff
@@ -83,3 +83,35 @@ fn test_translationtable_translate() {
         "This h is a helicopter!".to_owned(),
     )
 }
+
+// Tries to guess the encoding, not perfect but does it's job
+pub fn decode_bytes<Bytes: Into<Vec<u8>>>(bytes: Bytes) -> Option<String> {
+    let bytes = bytes.into();
+    let (label, confidence, _) = chardet::detect(&bytes);
```
pickfire, line 90: I don't know if we need `chardet`, why not use `for_label` from `encoding_rs` to get the encoding?

charliethomson: `from_label` expects an encoding label, `chardet` gets us a label for the encoding

src/deroff.rs
charliethomson: Maybe post-mvp?

src/util.rs
pickfire: Is `from_label` a typo? I mean https://docs.rs/encoding_rs/0.8.23/encoding_rs/struct.Encoding.html#method.for_label, I thought it takes in the original data?

src/deroff.rs
pickfire: Yes, this could be post-mvp but I think it still needs not to do unnecessary cloning that is very likely useless.

src/util.rs
scooter-dangle: I haven't used that method before, but yeah, it _looks_ like it's just the parser logic for a label, which we won't have.

charliethomson: Yes `from_label` was a typo, `for_label` expects the label, not the data, `detect` gets the label from the data

charliethomson:
post-mvp

Issue #105, opened 2020-08-03
Match feature additions to Python source since our snapshot


scooter-dangle:
To get the diff yourself, run the following from within the https://github.com/fish-shell/fish-shell repo
```sh
git diff --ignore-all-space e7bfd1d71ca54df726a4f1ea14bd6b0957b75752 master share/tools/create_manpage_completions.py
```

Many of the changes are strictly formatting (e.g., quoting strings with `"` rather than `'`). Substantive changes are those such as the following: https://github.com/fish-shell/fish-shell/blob/2083acec2e7fb6432ddf09c64f703e74c91b2d95/share/tools/create_manpage_completions.py#L263

There's also the addition of the `TypeScdocManParser` class.

I've omitted the diff for share/tools/deroff.py, as those changes appear to be exclusively stylistic or otherwise irrelevant to us.

main.py changes:

```diff
diff --git a/share/tools/create_manpage_completions.py b/share/tools/create_manpage_completions.py
index bbee8d102..97e48e49a 100755
--- a/share/tools/create_manpage_completions.py
+++ b/share/tools/create_manpage_completions.py
@@ -16,8 +16,19 @@ Redistributions in binary form must reproduce the above copyright notice, this l
[...]

PR #106, opened 2020-08-05
feat(main): tackling the final 😈 (WIP)

scooter-dangle:


pickfire:
Oh, unicode branch name. I didn't know branch name could be unicode.

pickfire:
I just realized most of the parsers are not working, I fixed `DeroffParser` which have some bugs where `take_while` created some issue when taking the current item, took me quite some time to fix.

scooter-dangle:
Copied from Zulip:

Hmm. When looking at how we handle fixtures/mlterm.1, seems to be 2 distinct bugs.

1) When I use the `--deroff-only` flag, I get output (using `--stdout`) that looks like the _actual_ completion outputs. This is due to the `deroff_only` field only triggering an explicit selection of the `ManType::TypeDeroff` variant (https://github.com/rust-dc/fish-manpage-completions/blob/43844c38c9404c5997a71782930539ba0a6b92d7/src/main.rs#L1315) but then not cutting off things off after doing the deroff step.

2) Excluding the `--deroff-only` flag helps to illustrate that we have another issue, I think, which is that something in our port of the conditional logic to determine which `ManType` isn't correctly getting us to `ManType::TypeDeroff`. I haven't checked yet to see which of the other `ManType`s that precede it in `ManType::ALL` is being selected.

pickfire:
> 1. When I use the --deroff-only flag, I get output (using --stdout) that looks like the actual completion outputs. This is due to the deroff_only field only triggering an explicit selection of the ManType::TypeDeroff variant (

That's because I double check and fixed the bugs in `TypeDeroff`, I didn't check the other deroffers yet, let me check it and fix the other bugs.

> 2. Excluding the --deroff-only flag helps to illustrate that we have another issue, I think, which is that something in our port of the conditional logic to determine which ManType isn't correctly getting us to ManType::TypeDeroff. I haven't checked yet to see which of the other ManTypes that precede it in ManType::ALL is being selected.

Fixed in https://github.com/rust-dc/fish-manpage-completions/pull/106/commits/b90a3d25b170571b36407f442dcbc3fba3e9dd25

pickfire:
Oh, I just realized original python `Type2` and `Type3` will always fail even though it runs successfully, I don't even know why it is there.

pickfire:
Now I am also planning to match new features in upstream https://github.com/rust-dc/fish-manpage-completions/issues/105. CC @charliethomson

src/main.rs
```diff
@@ -520,9 +529,18 @@ fn remove_groff_formatting(data: &str) -> Cow<str> {
...
-    let re =
-        regex!(r"\\fI|\\fP|\\f1|\\fB|\\fR|\\e|\.BI|\.BR|0\.5i|\.rb|\\\^|\{ | \}|\.B|\.I|.PD( \d+)");
-    re.replace_all(&data, "")
+    let re1 = regex!(
+        r"\\fI|\\fP|\\f1|\\fB|\\fR|\\e|\.BI|\.BR|0\.5i|\.rb|\\\^|\{ | \}|\.B|\.I|\f|(.PD( \d+))"
+    );
+    let re2 = regex!(r"\\-");
+    let re3 = regex!(r"\(cq");
+    match re1.replace_all(&data, "") {
+        Cow::Borrowed(s) => match re2.replace_all(&s, "-") {
```
scooter-dangle, line 538: :cow: :+1:

src/main.rs
```diff
@@ -351,17 +364,14 @@ fn truncated_description(description: &str) -> String {
         .split(".")
         .filter(|sentence| !sentence.trim().is_empty());
 
-    let mut out = format!(
-        "{}.",
-        lossy_unicode(&sentences.next().unwrap_or_default().as_bytes())
-    );
+    let out = sentences.next().unwrap_or_default().trim_end_matches('.');
```
scooter-dangle, line 367: Why remove the trailing `.`? Was this for a particular test case? Some of the changes in the function cause the first assertion of `test_truncated_description` to fail.

src/main.rs
```diff
@@ -351,17 +364,14 @@ fn truncated_description(description: &str) -> String {
...
+    let mut out = lossy_unicode(out.as_bytes());
     let mut out_len = char_len(&out);
 
     if out_len > MAX_DESCRIPTION_WIDTH {
         out = char_truncate_string(&out, MAX_DESCRIPTION_WIDTH, TRUNCATION_SUFFIX).into_owned();
     } else {
         for line in sentences {
-            let line = lossy_unicode(&line.as_bytes());
```
scooter-dangle, line 364: Is it accurate to remove this function call? It exists here in the original Python source.

src/main.rs
pickfire: Oh, I took it from upstream python but I forgot to update the tests. I was wondering why upstream python does not have the trailing `.` and noticed that ours is probably a bit old. I fixed the behavior to only remove the last period rather than the first one.

src/main.rs
pickfire: If I am correct, we already run `lossy_unicode` when we first loads the file so I think it is unnecessary to do it again. I think it was also called above but I think that could also be removed.

pickfire: Oh, you are correct. We need to run it since it was not runned for the other parts of the description.

pickfire: I think we could also remove the `lossy_unicode` since it is always allocating a `String`, we don't need that when we do only `format!` for those usage, I use `String::from_utf8_lossy(s.as_bytes())` instead.

PR #107, opened 2020-08-05
Tentative deroff test

charliethomson:
There are still a lot of failures, need to figure out why. Currently the test outputs the `get_output()` result of the deroffer to `./test_deroff_output/<manpage_name>.deroffed`, but only if the deroffed was different from the deroffed from the python. It also clones the `manpages` repo (so we have a consistent test suite). There are lots of bugs, somebody please help me find them :sob:
It looks like the majority of the failures are due to the `.de` macro, or `.nr`, but please investigate, i've run out of brain to diag 

🦀

### DIFF
--- a/src/deroff.rs
+++ b/src/deroff.rs
@@ -1096,7 +1096,7 @@ impl Deroffer {
             .next()
             .expect("`do_line` called when `self.s` was empty")
         {
-            b'.' | b'\'' => !self.request_or_macro(),
+            b'.' | b'\'' => self.request_or_macro(),
             _ => {
                 if self.tbl {
                     self.do_tbl()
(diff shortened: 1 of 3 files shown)


src/deroff.rs
```diff
@@ -1477,12 +1484,212 @@ fn test_request_or_macro() {
...
+    fs::create_dir(TEST_OUTPUT_DIR)?;
+
+    let stdout = io::stdout();
+    let mut lock = stdout.lock();
+
+    // If we don't have the manpages, get them
+    if !Path::new("./manpages").exists() {
+        print!("Cloning remote manpages... ");
+        lock.flush()?;
+        match Repository::clone("https://github.com/rust-dc/manpages.git", "./manpages") {
```
pickfire, line 1515: I don't think we should be doing this in our test. Maybe we could clone it first as a submodule.

pickfire:
@charliethomson I haven't check this yet but how do you want me to help?

pickfire:
I just realized I have written quite a bit of bugs here and there. Some fixed in https://github.com/rust-dc/fish-manpage-completions/pull/106 but still cannot output to standard output yet.

scooter-dangle:
@charliethomson, do we still want to merge this? I get conflicts and then build errors when I try to resolve the conflicts locally.

PR #108, opened 2020-12-20
feat(main): add logging

pickfire:
Added logging with tracing, did not use log because log does not have span support.

Next step is to switch to indicatif and rayon (and option to not use threads or set threads) to display multiple progress bar while doing multiple stuff.

Logging is unlike the old logging which uses indentation, using indentation is not a good idea when we want to go parallel later (stuff gets mixed up), we could keep a buffer but that makes stuff complicated.

Screenshot_20201221_000643

Lately got lazy so this patch came in slow.


src/main.rs
```diff
@@ -1366,13 +1247,15 @@ fn parse_manpage_at_path(
...
+        tracing::info!("{}: Not supported", manpage_path.display());
     }
 
-    if let Some(mut completions) = parsers
-        // .inspect(|parser| add_diagnostic(format!("Trying {}", parser)))
-        .find_map(|parser| parser.parse_man_page(&manpage, &cmdname))
-    {
+    if let Some(mut completions) = parsers.find_map(|parser| {
+        // Second (last) level span
+        let span = tracing::info_span!("Trying", "{}", parser.to_string().as_str());
```
scooter-dangle, line 1255:
```suggestion
        let span = tracing::info_span!("Trying", "{}", parser);
```

src/main.rs
```diff
@@ -1305,13 +1186,13 @@ fn parse_manpage_at_path(
...
-
-    // Set up some diagnostic
-    // add_diagnostic(format!("Considering {}", manpage_path));
-    // diagnostic_indent += 1
+    // First level span
+    let span = tracing::info_span!(
+        "Considering",
+        "{}",
+        manpage_path.to_string_lossy().into_owned().as_str()
+    );
```
scooter-dangle, line 1194:
```suggestion
    let span = tracing::info_span!("Considering", "{}", manpage_path.display());
```

scooter-dangle (changes requested):

src/main.rs
```diff
@@ -1634,6 +1517,14 @@ fn program_name() -> String {
 fn main() -> Result<(), String> {
     let opts = Opts::from_args();
 
+    let level = match opts.verbose {
+        0 => LevelFilter::OFF,
+        1 => LevelFilter::WARN,
+        2 => LevelFilter::INFO,
+        n => unreachable!(n),
+    };
+    tracing_subscriber::fmt().with_max_level(level).init();
```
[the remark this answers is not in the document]
pickfire: Thanks, I didn't notice this.

src/main.rs
```diff
@@ -11,67 +11,14 @@ use bzip2::read::BzDecoder;
...
 
-// # -*- coding: utf-8 -*-
-//
-// # Run me like this: ./create_manpage_completions.py /usr/share/man/man{1,8}/* > man_completions.fish
-//
-// """
-// <OWNER> = Siteshwar Vashisht
-// <YEAR> = 2012
-//
-// Copyright (c) 2012, Siteshwar Vashisht
```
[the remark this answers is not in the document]
pickfire: I already did the previous one, is this still valid?

src/main.rs
pickfire: Nice catch, I didn't check the blocks a lot when I was doing modification.

scooter-dangle (approved):
Awesome. Thanks! 🎉

pickfire:
@scooter-dangle T_T I had another cleanup commit

scooter-dangle:
> @scooter-dangle T_T I had another cleanup commit

Womp womp. My bad! Tiny cleanup PR?

PR #110, opened 2021-05-27
fix(drain): panic when different length

pickfire:
Python slice behavior works even when the end length is not achieved
but just the output length will be shorter but rust length requires it
to hit exactly that length and when the length is not enough or within
a boundary, it will panic. So we have to count by character boundary
and at the same time prevent the characters length from overflowing.

I hit some panics when running it through my system files when I wanted to test out the progress bar (done but still testing).

### DIFF
--- a/src/deroff.rs
+++ b/src/deroff.rs
@@ -358,7 +358,15 @@ impl Deroffer {
     }
 
     fn skip_char(&mut self, amount: usize) {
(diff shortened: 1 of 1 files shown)


scooter-dangle (approved):
El jee tea em, thanks! :+1: :tada:

Issue #82, opened 2020-02-29
Progress bar


pickfire:
Should we do something like cargo which is quite verbose especially we have small tasks (I have ~10000 man pages to be generated)? Or should we use something like [indicatif](https://crates.io/crates/indicatif) which is able to be used with rayon (see also #50)?

charliethomson:
I looked into indicatif the other day, it looks really good. definitely has my vote

scooter-dangle:
Resolved by #111

PR #111, opened 2021-05-27
feat: add basic progress bar

### LINKED ISSUE #50: Parallelize completion generation
Once we have a working MVP, we should address the opportunities for embarrassingly parallel optimization.

pickfire:
Add progress bar with indicatif while parallelizing stuff.
It does not show what file it is processing so we may need to have a
custom painter.
Fix #50, #82

Debug ~20s
Release ~600ms

output-opt

A simple implementation for this does not show custom progress, but if we want we need to do some custom painting, it can be done later.

While working on this I found logging missing some stuff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,6 +170,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
+version = "0.14.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3993e6445baa160675931ec041a5e03ca84b9c6e32a056150d3aa2bdda0a1f45"
+dependencies = [
+ "encode_unicode",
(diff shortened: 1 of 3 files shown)


scooter-dangle:
@pickfire, hey!

Were you thinking of doing anything more on this branch or should I just cleanup some of the commented-out code and merge it in?

- - -

I knew we could parallelize, but jeez, rayon gives us a _massive_ kick in the pants. If I run the current `fish_update_completions` locally, it takes about 16 seconds. For the same inputs, the release build on this PR comes in at just under 400ms.

scooter-dangle (approved):
Looks great! Thanks!

pickfire:
No, I forgot what was I thinking for the next step so need to check back again.

PR #113, opened 2021-11-16
fix: add missing fallback path from original Python source

scooter-dangle:
See
https://github.com/fish-shell/fish-shell/blob/15a3caf244cf62b7a4342481760b1538f5f0b655/share/tools/create_manpage_completions.py#L1003

Maybe there was a reason for excluding this particular path?

@pickfire, not sure why we would have excluded this particular path. Maybe something I'm missing?


pickfire (approved):
Looks good to me.

pickfire:
Probably I missed it previously.

PR #135, opened 2023-02-22
Bump dependencies

pickfire:


pickfire:
@scooter-dangle pipeline issue

scooter-dangle:
Hmm. Might be. Kinda surprised at the error message if it's due to the image being old, since you didn't change the way we call `cargo fmt`. Re-running the job.

src/main.rs
```diff
@@ -1504,34 +1496,35 @@ fn program_name() -> String {
...
     tracing_subscriber::fmt()
         .with_writer(io::stderr)
         .with_max_level(level)
         .init();
 
-    if opts.completions {
-        Opts::clap().gen_completions_to(
+    if args.completions {
+        clap_complete::generate(
+            clap_complete::Shell::from_env().expect("Unsupported shell"),
```
scooter-dangle, line 1514: Ah, nice. Glad we don't have to do the shell lookup manually anymore.

src/main.rs
```diff
@@ -1450,51 +1450,43 @@ fn get_paths_from_man_locations() -> Vec<PathBuf> {
...
-#[derive(StructOpt, Debug)]
-struct Opts {
+#[derive(Parser, Debug)]
+struct Args {
     /// Level of debug output.
-    #[structopt(short, long, default_value = "0", possible_values = &["0", "1", "2"])]
+    #[clap(
+        short, long, default_value = "0",
+        value_parser = clap::value_parser!(u8).range(0..3)
+    )]
```
scooter-dangle, line 1459: Nice! Is this evaluated at runtime or compile-time?

src/main.rs
```diff
@@ -1504,34 +1496,35 @@ fn program_name() -> String {
...
-    let opts = Opts::from_args();
+    let args = Args::parse();
 
-    let level = match opts.verbose {
+    let level = match args.verbose {
         0 => LevelFilter::OFF,
         1 => LevelFilter::WARN,
         2 => LevelFilter::INFO,
-        n => unreachable!(n),
+        n => unreachable!("{}", n),
```
scooter-dangle, line 1505:
```suggestion
        n => unreachable!("{n}"),
```

scooter-dangle (approved):

scooter-dangle:
not sure what was up, but it works with a more recent build of the docker image

src/main.rs
pickfire: Compile time I believe, but the error for this is kinda bad, I wish it is `[0, 1, 2]` but I didn't look closely.

